PR #2, opened 2020-08-26
Add some missing model names

bieniu:
This PR:
 - adds some missing model names (models with relays)
 - sorts model names


balloob (approved):

balloob:
Thanks!

PR #4, opened 2020-08-27
Add supported devices list

Misiu:
based on https://github.com/home-assistant-libs/aioshelly/blob/main/aioshelly/__init__.py#L8-L17

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,17 @@ if __name__ == "__main__":
     asyncio.run(main())
 ```
 
+## Supported devices
+
+- Shelly 1
+- Shelly 1 PM
+- Shelly 2 (relay mode for now)
+- Shelly 2.5 (relay mode for now)
+- Shelly 4 Pro
+- Shelly H&T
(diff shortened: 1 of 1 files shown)


balloob:
Those are not supported, those are just ones that we have human readable names for.

balloob:
Every Shelly device is supported. We can read their info and blocks and allow `set_state`.

Misiu:
@balloob if every Shelly device is supported I can extend the list.
What about roller shutter mode for Shelly 2.5?
Should I ask for support here or in HA repo?
This mode is most useful for me.

balloob:
This wrapper shows the data from the API and that's it. It doesn't care if something is in roller mode or not. That's up to the something that uses this lib to implement. So not this repo.

Also no feature requests in the core repo but on the forums 👍

Misiu:
@balloob thanks for the reeply.
I'll create FR on the forum and will test new integration with devices I have.

PR #13, opened 2020-09-12
Improved logging (and added "id" property)

chemelli74:


aioshelly/__init__.py
```diff
@@ -257,7 +261,7 @@ def __init__(
 
             if sensor[BLOCK_VALUE_TYPE] != BLOCK_VALUE_TYPE_TEMPERATURE:
                 raise ValueError(
-                    "Found duplicate description for non-temperature sensor"
+                    f"Found duplicate description for non-temperature sensor for device {self.device.ip} [{self.device.id}]"
```
balloob, line 264: We can't use `id`. It's not available when we have no authentication and so would cause a second exception to be raised.

chemelli74: As the `ValueError` is inside the if with a temperature check, we should already be authenticated at this point.

Simone

balloob: That's not true. Authentication is only required for the HTTP requests. Blocks are fetched from the COAP API and do not require authentication.

chemelli74: So no way to improve logging; as currently is very hard to get the issue when you have 30+ devices

balloob: I think that you're trying to solve the issue at the wrong place? The library raises accordingly, it's up to the consumer to report the issue and attribute to where it's used. Also, the example script can always be used to easily reproduce.

chemelli74: I accept your decision, but last time we experienced it, was a missing guard check causing it and example script was not failing.
When I develop, I always take particular care of detailed error messages because they help user fix his local issue and instead if it's a code issue they help us identify it quickly and with less effort.

Simone

PR #14, opened 2020-09-12
Fix channel property when description doesn't contain "_"

bieniu:
For some devices, the sensors are in the block with the description `device` and the `channel` property generates an error:
```
  File "/home/maciek/develop/home-assistant-core/homeassistant/helpers/entity_platform.py", line 193, in _async_setup_platform
    await asyncio.shield(task)
  File "/home/maciek/develop/home-assistant-core/homeassistant/components/shelly/sensor.py", line 116, in async_setup_entry
    await async_setup_entry_attribute_entities(
  File "/home/maciek/develop/home-assistant-core/homeassistant/components/shelly/entity.py", line 48, in async_setup_entry_attribute_entities
    [
  File "/home/maciek/develop/home-assistant-core/homeassistant/components/shelly/entity.py", line 49, in <listcomp>
    sensor_class(wrapper, block, sensor_id, description, counts[block.type])
  File "/home/maciek/develop/home-assistant-core/homeassistant/components/shelly/entity.py", line 148, in __init__
    name_parts.append(str(block.channel))
  File "/home/maciek/develop/home-assistant-core/venv/lib/python3.8/site-packages/aioshelly/__init__.py", line 280, in channel
    return self.description.split("_")[1]
IndexError: list index out of range
```
This PR fixes this.

### DIFF
--- a/aioshelly/__init__.py
+++ b/aioshelly/__init__.py
@@ -275,7 +275,7 @@ class Block:
 
     @property
     def channel(self):
-        return self.description.split("_")[1]
+        return self.description.split("_")[1] if "_" in self.description else None
 
     def info(self, attr):
         """Return info over attribute."""


chemelli74:
Which devices are showing this behavior ?

bieniu:
I'm sure about Shelly 2 (SHSW-21) but I think the same problem concerns Shelly 1PM (SHSW-PM) and Shelly 2.5 (SHSW-25).

chemelli74:
> I'm sure about Shelly 2 (SHSW-21) but I think the same problem concerns Shelly 1PM (SHSW-PM) and Shelly 2.5 (SHSW-25).

I'm running both 1PM and 2.5 and cannot see this issue.
Referencing https://shelly-api-docs.shelly.cloud/docs/coiot/v2/examples, if I don't get it wrong also 2 should work with current code.

Am I completely on a wrong track ?

bieniu:
For every Shelly device this block description doesn't contains `_`.
This is an example for Shelly 2.5:
```
    "blk":[
        {
            "I":1,
            "D":"relay_0"
        },
        {
            "I":2,
            "D":"relay_1"
        },
        {
            "I":4,
            "D":"device"
        }
    ],
```

chemelli74:
> For every Shelly device this block description doesn't contains `_`.
> This is an example for Shelly 2.5:
> 
> ```
>     "blk":[
>         {
>             "I":1,
>             "D":"relay_0"
>         },
>         {
>             "I":2,
>             "D":"relay_1"
>         },
>         {
>             "I":4,
>             "D":"device"
>         }
>     ],
> ```

both channel descriptions have the underscore:

- relay_0
- relay_1

isn't it ?

Simone

bieniu:
What about `device`?

chemelli74:
> What about `device`?

Indeed, but device has no info about channels, so `block.channel` should not be called for device but only for relay/cover/lights.

Simone

bieniu:
> device has no info about channels

So `channel` property for `device` block should return `None`. Now it raises an exception.

chemelli74:
I suppose you have local changes.
What are you using device.channel for ?

bieniu:
> I suppose you have local changes.

Nope, I have this error on 0.115.0b5 with Shelly 2 for `energy` and `power` sensors.
We use `channel` [here](https://github.com/home-assistant/core/blob/dev/homeassistant/components/shelly/entity.py#L148) but I have to find out why `same_type_count is greater than 1.

chemelli74:
Check latest commit from Paulus on Shelly integration (core).
I think check is done only for Shelly 2.5, Shelly 2 is missing

balloob (approved):
This PR is fine. It's indeed None.

balloob:
If there are 2 of the same type, it has to have a channel 🤔 that's a bug in HA

chemelli74:
> > I suppose you have local changes.
> 
> Nope, I have this error on 0.115.0b5 with Shelly 2 for `energy` and `power` sensors.
> We use `channel` [here](https://github.com/home-assistant/core/blob/dev/homeassistant/components/shelly/entity.py#L148) but I have to find out why `same_type_count is greater than 1.

Something weird if you have more than a device block for a device.
Cannot reproduce with any of the device I have (1, 1pm, 2.5, plug-s, em, door, flood, h&t, air, gas); but I miss Shelly 2.

Simone

PR #16, opened 2020-09-19
Implemented method for REST updates

chemelli74:


bieniu (commented):
Do we need update `settings` and `status` at a time? `status` will be updated periodically, `settings` on init only.

chemelli74:
> Do we need update `settings` and `status` at a time? `status` will be updated periodically, `settings` on init only.

Settings can be changed as well, not as frequently as status but still.

bieniu:
But we only use data from settings during initialization. I think we need the `update_status` method, and if we ever need to update the settings, we will add the `update_settings` method.

chemelli74:
> But we only use data from settings during init. I think we need the `update_status` method, and if we ever need to update the settings, we will add the `update_settings` method.

Done.
We can also consider using "cfg_changed_cnt" from status to know if a call to update settings is needed or not.

bieniu:
That's a good idea. We have to consider whether we are interested in changing the device configuration during HA operation.

chemelli74:
> That's a good idea. We have to consider whether we are interested in changing the device configuration during HA operation.

Let's keep it for a later use, if needed.

balloob:
Would be cool if we would automatically update settings if status contains a new config count. We should in that case rename the function name, as it's not just status that is being updated.

PR #17, opened 2020-09-19
Make code more pylint compliant

chemelli74:
Resolved many warnings, still getting some:

```
************* Module aioshelly.__init__
__init__.py:98:0: C0330: Wrong hanging indentation before block (add 4 spaces).
        f"http://{ip_address}/shelly", raise_for_status=True
        ^   | (bad-continuation)
__init__.py:107:0: C0330: Wrong hanging indentation before block (add 4 spaces).
        self,
        ^   | (bad-continuation)
__init__.py:108:0: C0330: Wrong hanging indentation before block (add 4 spaces).
        coap_context: aiocoap.Context,
        ^   | (bad-continuation)
__init__.py:109:0: C0330: Wrong hanging indentation before block (add 4 spaces).
        aiohttp_session: aiohttp.ClientSession,
        ^   | (bad-continuation)
__init__.py:110:0: C0330: Wrong hanging indentation before block (add 4 spaces).
        options: ConnectionOptions,
        ^   | (bad-continuation)
__init__.py:124:0: C0330: Wrong hanging indentation before block (add 4 spaces).
        cls, aiohttp_session, ip_or_options: Union[str, ConnectionOptions]
        ^   | (bad-continuation)
__init__.py:262:0: C0330: Wrong hanging indentation before block (add 4 spaces).
        self, device: Device, blk_type: str, blk: dict, sensors: Dict[str, dict]
        ^   | (bad-continuation)
************* Module aioshelly
__init__.py:103:0: R0902: Too many instance attributes (9/7) (too-many-instance-attributes)
__init__.py:193:17: E1101: Module 'aiocoap' has no 'GET' member (no-member)

```

PR #19, opened 2020-09-23
Fix parameter names after PR#17

bieniu:
We have a few parameter name issues after #17 
This PR fixes these mistakes.


balloob:
@chemelli74 you need to test your PRs! These fixes show that you didn't run your PR. We need to do better.

chemelli74:
Sorry if we did a too quick test. Will do better in future.

Simone

PR #36, opened 2020-10-25
Example.py: introduction of devices external list

chemelli74:


example.py
```diff
@@ -39,10 +40,17 @@ async def cli():
 
 
 async def test_many():
-    device_options = [
-        aioshelly.ConnectionOptions("192.168.1.165", "admin", "test-password"),
-        aioshelly.ConnectionOptions("192.168.1.168"),
-    ]
+    with open("devices.json") as j:
+        json_data = json.load(j)
+    print(json_data)
```
balloob, line 45: unnecessary print.

devices.json
```diff
@@ -0,0 +1 @@
+{"devices": [{"ip": "192.168.1.1", "username": "admin", "password": "password"}, {"ip": "192.168.1.2"}, {"ip": "192.168.1.3"}]}
```
balloob, line 1: I would much rather just have 1 device defined per line, and we load 1 line at a time into JSON.

example.py
chemelli74: yup, leftover

example.py
```diff
@@ -39,10 +40,16 @@ async def cli():
 
 
 async def test_many():
-    device_options = [
-        aioshelly.ConnectionOptions("192.168.1.165", "admin", "test-password"),
-        aioshelly.ConnectionOptions("192.168.1.168"),
-    ]
+    with open("devices.json") as j:
```
balloob, line 43: add this file to `.gitignore`

chemelli74: I though to have one in the repo as sample for users

balloob: We can do both. Add a default, then add it to gitignore to ignore future updates.

example.py
```diff
@@ -43,10 +44,16 @@ def device_updated(device):
...
+    device_options = []
+    with open("devices.json") as fp:
+        for line in fp:
+            obj = json.loads(line)
+            device_data = []
+            if "username" in obj:
+                device_data = [obj["ip"], obj["username"], obj["password"]]
+            else:
+                device_data = [obj["ip"]]
+            device_options.append(aioshelly.ConnectionOptions(device_data[0]))
```
balloob, line 56: You're not using user/pass defined in `device_data`. There is also no reason to define `device_data` as intermediary. You can just read from `obj`.

balloob (approved):

PR #38, opened 2020-10-26
Fix for no CoAP context (verify.py)

chemelli74:


aioshelly/__init__.py
```diff
@@ -144,9 +144,10 @@ def __init__(
         self._settings = None
         self.shelly = None
         self._status = None
-        self._unsub_listening = coap_context.subscribe_updates(
-            options.ip_address, self._coap_message_received
-        )
+        if coap_context:
```
balloob, line 147: This is wrong. We always need a coap context or else you cannot fetch cit/s and cit/d.

chemelli74: verify.py script in this repo doesn't use a coap context because it doesn't fetch live data but just validate vs json file of cit/d and cit/s.

balloob: Then the verify script needs to pass in a mock object. We should not allow setting up our class without a coap object as it will break.

balloob:
```python
from unittest.mock import Mock

COAP = MocK()
```

chemelli74: but that way a call to `coap_context.subscribe_updates()` still fails

balloob: Push up the code that you have.

balloob:
da2c8d0a75cdd0afe4b81c95996b229936879eb5 👀

chemelli74:
> [da2c8d0](https://github.com/home-assistant-libs/aioshelly/commit/da2c8d0a75cdd0afe4b81c95996b229936879eb5) 👀

Coffe was not enough to wake up today :-(

I was looking at device `create()` and not `__init__()`.

Simone

PR #39, opened 2020-10-27
Use datagram transport to send data + fix mypy

balloob:
- Add mypy to CI
- Use datagram transport to send data so it's non blocking


aioshelly/coap.py
```diff
@@ -60,13 +62,18 @@ async def request(self, ip: str, path: str):
...
         Subscribe with `subscribe_updates` to receive answer.
         """
+        assert self.transport is not None
         msg = b"\x50\x01\x00\x0A\xb3cit\x01" + path.encode() + b"\xFF"
-        self.sock.sendto(msg, (ip, 5683))
+        self.transport.sendto(msg, (ip, 5683))
 
     def close(self):
         """Close."""
         self.sock.close()
```
MartinHjelmare, line 71: Do we need to close the transport or is closing the socket enough?

balloob: Oh good question. Looked it up, the DatagramTransport inherits from the BaseTransport which [does not implement close](https://github.com/python/cpython/blob/master/Lib/asyncio/transports.py#L27-L35)

MartinHjelmare: Wonder where the code is that does what it says it should do?

https://docs.python.org/3/library/asyncio-protocol.html#asyncio.BaseTransport.close

balloob: Not in Python 😆 

```
>>> import asyncio
>>> dg = asyncio.DatagramTransport()
>>> dg.close()
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.8/asyncio/transports.py", line 35, in close
    raise NotImplementedError
NotImplementedError
```

chemelli74: > https://docs.python.org/3/library/asyncio-protocol.html#asyncio.BaseTransport.close

https://github.com/python/cpython/blob/a6879d9445f98833c4e300e187956e2a218a2315/Lib/asyncio/transports.py#L27

MartinHjelmare: Maybe this is overwritten somehow by the socket or in the c code?

The docs example server and client works:
```
Python 3.8.5 (default, Jul 28 2020, 12:59:40) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.18.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: 

In [1]: import asyncio
   ...: 
   ...: 
   ...: class EchoClientProtocol:
   ...:     def __init__(self, message, on_con_lost):
   ...:         self.message = message
   ...:         self.on_con_lost = on_con_lost
   ...:         self.transport = None
   ...: 
   ...:     def connection_made(self, transport):
   ...:         self.transport = transport
   ...:         print('Send:', self.message)
   ...:         self.transport.sendto(self.message.encode())
   ...: 
   ...:     def datagram_received(self, data, addr):
   ...:         print("Received:", data.decode())
   ...: 
   ...:         print("Close the socket")
   ...:         self.transport.close()
   ...: 
   ...:     def error_received(self, exc):
   ...:         print('Error received:', exc)
   ...: 
   ...:     def connection_lost(self, exc):
   ...:         print("Connection closed")
   ...:         self.on_con_lost.set_result(True)
   ...: 
   ...: 
   ...: async def main():
   ...:     # Get a reference to the event loop as we plan to use
   ...:     # low-level APIs.
   ...:     loop = asyncio.get_running_loop()
   ...: 
   ...:     on_con_lost = loop.create_future()
   ...:     message = "Hello World!"
   ...: 
   ...:     transport, protocol = await loop.create_datagram_endpoint(
   ...:         lambda: EchoClientProtocol(message, on_con_lost),
   ...:         remote_addr=('127.0.0.1', 9999))
   ...: 
   ...:     try:
   ...:         await on_con_lost
   ...:     finally:
   ...:         transport.close()
   ...: 
   ...: 
   ...: asyncio.run(main())
Send: Hello World!
Received: Hello World!
Close the socket
Connection closed
```

balloob: Found it ! https://github.com/python/cpython/blob/a6879d9445f98833c4e300e187956e2a218a2315/Lib/asyncio/selector_events.py#L87

(by using `print(transport.close.__code__)`

aioshelly/coap.py
```diff
@@ -60,12 +62,19 @@ async def request(self, ip: str, path: str):
...
+        assert self.transport is not None
         msg = b"\x50\x01\x00\x0A\xb3cit\x01" + path.encode() + b"\xFF"
-        self.sock.sendto(msg, (ip, 5683))
+        self.transport.sendto(msg, (ip, 5683))
 
     def close(self):
         """Close."""
-        self.sock.close()
+        print("CLOSING")
+        print(self.transport.close.__code__)
```
MartinHjelmare, line 72:
```suggestion
```

MartinHjelmare (approved):
Looks good!

aioshelly/coap.py
balloob: 🤦

PR #49, opened 2020-12-06
Use gethostbyname to resolve hostname to IP address

bieniu:
The CoAP listener needs the device's IP address, so with this change, `aioshelly` will use `socket.gethostbyname` to resolve hostname to IP address. This changed was suggested here https://github.com/home-assistant/core/pull/43963

### DIFF
--- a/aioshelly/__init__.py
+++ b/aioshelly/__init__.py
@@ -162,10 +162,14 @@ class Device:
         ip_or_options: Union[str, ConnectionOptions],
     ):
         """Device creation."""
+        loop = asyncio.get_running_loop()
         if isinstance(ip_or_options, str):
-            options = ConnectionOptions(gethostbyname(ip_or_options))
+            ip_addr = await loop.run_in_executor(None, gethostbyname, ip_or_options)
+            options = ConnectionOptions(ip_addr)
         else:
-            ip_or_options.ip_address = gethostbyname(ip_or_options.ip_address)
+            ip_or_options.ip_address = await loop.run_in_executor(
+                None, gethostbyname, ip_or_options.ip_address
+            )
             options = ip_or_options
 
         instance = cls(coap_context, aiohttp_session, options)
--- a/aioshelly/__init__.py
+++ b/aioshelly/__init__.py
@@ -163,7 +163,6 @@ class Device:
         ip_or_options: Union[str, ConnectionOptions],
     ):
         """Device creation."""
-        loop = asyncio.get_running_loop()
         if isinstance(ip_or_options, str):
             options = ConnectionOptions(ip_or_options)
         else:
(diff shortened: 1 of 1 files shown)


aioshelly/__init__.py
```diff
@@ -162,8 +163,9 @@ async def create(
     ):
         """Device creation."""
         if isinstance(ip_or_options, str):
-            options = ConnectionOptions(ip_or_options)
+            options = ConnectionOptions(gethostbyname(ip_or_options))
```
balloob, line 166: Is this doing I/O ? 

Also, should it be done in this method, or done when we create a coap context which actually needs it?

bieniu: > Is this doing I/O ?

Hmm I think, yes. I'm not sure what to do with that. Call it in the event loop?

> Also, should it be done in this method, or done when we create a coap context which actually needs it?

If we resolve hostname to IP only in `COAP.subscribe_updates()` then the aioshelly device `ip_address` property will contain the hostname, not IP. We can use `gethostbyname` in `Device.__init__()`
```python
class Device:
    """Shelly device reppresentation."""

    def __init__(
        self,
        coap_context: COAP,
        aiohttp_session: aiohttp.ClientSession,
        options: ConnectionOptions,
    ):
        options.ip_address = gethostbyname(options.ip_address)
        self.coap_context = coap_context
        self.aiohttp_session = aiohttp_session
        self.options = options
```

balloob: Well first find out if it does I/O. If it does, it should run in the executor.

bieniu: `gethostbyname` does I/O. Can we proceed with this change?

chemelli74 (approved):

aioshelly/__init__.py
```diff
@@ -162,10 +162,14 @@ async def create(
         ip_or_options: Union[str, ConnectionOptions],
     ):
         """Device creation."""
+        loop = asyncio.get_running_loop()
         if isinstance(ip_or_options, str):
-            options = ConnectionOptions(gethostbyname(ip_or_options))
+            ip_addr = await loop.run_in_executor(None, gethostbyname, ip_or_options)
```
balloob, line 167: Why not just add this resolving after the if…else check and do it only if it's not an IP?

chemelli74 (commented):
I see you applied Paulus suggestion, but to be honest I think that a try/exception is more expensive and less cleaner that always calling a `gethostbyname`.

IMHO, the latest is made to resolve ip and host for a reason ;-)

balloob (approved):

balloob:
@chemelli74 running things in the executor is expensive. It means a context switch to another thread. Try…Except really isn't that expensive.

PR #59, opened 2021-01-07
Enable multicast membership to be call as needed

chemelli74:


chemelli74:
Changed approach, closing

PR #62, opened 2021-01-18
Add support for initializing device when cit/s arrives

thecode:
Add `initialize` parameter to `Device.create`, default to `True`
If set to `False` will not initialize device on create but only after `cit/s` CoAP message arrives from the device.
Add new property `initialized` which set to `True` after initialization finished.
Raise `NotInitialized` exception when trying to get value from non-initialized device.
Migrate to `argparse`

### DIFF
--- a/example.py
+++ b/example.py
@@ -1,4 +1,5 @@
 # Run with python3 example.py <ip of shelly device>
+import argparse
 import asyncio
 import json
 import sys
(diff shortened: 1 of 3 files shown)


example.py
```diff
@@ -28,20 +30,21 @@ async def cli():
...
-                aioshelly.Device.create(aiohttp_session, coap_context, options), 5
+                aioshelly.Device.create(aiohttp_session, coap_context, options, init), 5
             )
         except asyncio.TimeoutError:
             print("Timeout connecting to", ip)
             return
 
-        print_device(device)
+        if device.initialized:
+            print_device(device)
```
balloob, line 40: I think that we should always print the device but teach `print_device` to show initializing devices.

example.py
```diff
@@ -12,13 +12,15 @@
...
     if len(sys.argv) < 2:
-        print("Error! Run with <ip> or <ip> <user> <pass>")
+        print("Error! Run with <ip> <init> or <ip> <init> <user> <pass>")
         return
 
     ip = sys.argv[1]
-    if len(sys.argv) > 3:
-        username = sys.argv[2]
-        password = sys.argv[3]
+    init = bool(sys.argv[2] in ["True", "true", "1"])
```
balloob, line 19: We should migrate to argparse https://github.com/home-assistant-libs/zwave-js-server-python/blob/master/zwave_js_server/__main__.py#L15-L30

example.py
thecode: Fixed

example.py
thecode: Migrated, thanks

balloob (approved):

chemelli74 (approved):

balloob (approved):

PR #65, opened 2021-01-29
Add multicast querier thread

chemelli74:
Unfortunately some routers/AP doesn't behave as multicast querier.

This leads to missing status query for the multicast group, thus making interested network devices (servers, client, etc.) to leave the group.
When HA server leaves the group, it stops receiving the CoAP status update packets loosing real time sync with Shelly devices.

This PR checks every 4 minutes ( longer interval in real world is 5 minutes ), if a multicast query packet is received and if not it sends one.

### DIFF
--- a/aioshelly/coap.py
+++ b/aioshelly/coap.py
@@ -1,17 +1,26 @@
 """COAP for Shelly."""
 import asyncio
+import ipaddress
 import json
 import logging
 import socket
 import struct
+import time
 from typing import Optional, cast
 
 import netifaces
-
-_LOGGER = logging.getLogger(__name__)
+from scapy.contrib.igmp import IGMP
+from scapy.layers.inet import IP
+from scapy.sendrecv import send, sniff
 
 MAIN_MULTICAST_IP = "224.0.1.187"
 WF200_MULTICAST_IP = "224.0.1.188"
+
+MULTICAST_QUERY_IGMPTYPE = 0x11
+MULTICAST_QUERY_GRP = "224.0.0.1"
+MULTICAST_QUERY_TIMEOUT = 240
+
+_LOGGER = logging.getLogger(__name__)
 
 
 class CoapMessage:
@@ -43,14 +52,20 @@ class CoapMessage:
             self.payload = None
 
 
-def get_all_ips():
-    """Get all ip from ethernet interfaces."""
-    ip_list = []
-    for iface in netifaces.interfaces():
+def get_network_objects(objtype: str):
+    """Get all ip/iface from system interfaces."""
+    obj_list = []
+    ifaces = netifaces.interfaces()
+    ifaces.remove("lo")
+
+    for iface in ifaces:
         iface_details = netifaces.ifaddresses(iface)
         if netifaces.AF_INET in iface_details:
-            ip_list.append(iface_details[netifaces.AF_INET][0]["addr"])
-    return ip_list
+            if objtype == "iface":
+                obj_list.append(iface)
+            else:
+                obj_list.append(iface_details[netifaces.AF_INET][0]["addr"])
+    return obj_list
 
 
 def socket_init():
@@ -59,7 +74,7 @@ def socket_init():
     sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
     sock.bind(("", 5683))
 
-    for ip in get_all_ips():
+    for ip in get_network_objects("ip"):
         for multicast_ip in MAIN_MULTICAST_IP, WF200_MULTICAST_IP:
             _LOGGER.debug("Adding ip %s to multicast %s membership", ip, multicast_ip)
             group = socket.inet_aton(multicast_ip) + socket.inet_aton(ip)
@@ -67,6 +82,49 @@ def socket_init():
 
     sock.setblocking(False)
     return sock
+
+
+class MulticastQuerier:
+    """Multicast querier management."""
+
+    def __init__(self):
+        """Initialize multicast querier thread."""
+        loop = asyncio.get_event_loop()
+        loop.run_in_executor(None, self.async_start)
+
+    @classmethod
+    def async_start(cls):
+        """Start sniffing for multicast query."""
+        filter_igmp_query = f"igmp and igmp[0] == {int(MULTICAST_QUERY_IGMPTYPE)} and (igmp[4:4] == {int(ipaddress.IPv4Address(MAIN_MULTICAST_IP))} or igmp[4:4] == 0)"
+        pkt_snd = IP(dst=MULTICAST_QUERY_GRP) / IGMP(
+            type=MULTICAST_QUERY_IGMPTYPE,
+            mrcode=100,
+            gaddr=MAIN_MULTICAST_IP,
+        )
+        while True:
+            result = sniff(
+                filter=filter_igmp_query,
+                store=1,
+                count=1,
+                timeout=MULTICAST_QUERY_TIMEOUT,
+            )
+            _LOGGER.debug("Multicast query sniff result: %s", result)
+            if not result:
+                _LOGGER.info(
+                    "Multicast query not received in %s seconds",
+                    MULTICAST_QUERY_TIMEOUT,
+                )
+                for eth_iface in get_network_objects("iface"):
+                    send(pkt_snd, iface=eth_iface, verbose=False)
+                    _LOGGER.info("Multicast query sent for %s interface", eth_iface)
+            else:
+                _LOGGER.info(
+                    "Multicast query received from network, no action required"
+                )
+            _LOGGER.debug(
+                "Multicast querier sleeping for %s seconds", MULTICAST_QUERY_TIMEOUT
+            )
+            time.sleep(MULTICAST_QUERY_TIMEOUT)
 
 
 class COAP(asyncio.DatagramProtocol):
@@ -85,6 +143,7 @@ class COAP(asyncio.DatagramProtocol):
         loop = asyncio.get_running_loop()
         self.sock = socket_init()
         await loop.create_datagram_endpoint(lambda: self, sock=self.sock)
+        MulticastQuerier()
 
     async def request(self, ip: str, path: str):
         """Request a CoAP message.
--- a/aioshelly/coap.py
+++ b/aioshelly/coap.py
@@ -7,6 +7,7 @@ import os
 import socket
 import struct
 import threading
+from ctypes.util import find_library
 from typing import Optional, cast
 
 import netifaces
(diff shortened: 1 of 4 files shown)


aioshelly/coap.py
```diff
@@ -1,18 +1,27 @@
...
 import logging
 import socket
 import struct
+import time
 from typing import Optional, cast
 
 import netifaces
-
-_LOGGER = logging.getLogger(__name__)
+from scapy.contrib.igmp import IGMP
```
balloob, line 12: Scapy requires root I think?

aioshelly/coap.py
```diff
@@ -43,14 +52,20 @@ def __init__(self, sender_addr, payload: bytes):
...
 
-def get_all_ips():
-    """Get all ip from ethernet interfaces."""
-    ip_list = []
-    for iface in netifaces.interfaces():
+def get_network_objects(objtype: str):
+    """Get all ip/iface from system interfaces."""
+    obj_list = []
+    ifaces = netifaces.interfaces()
+    ifaces.remove("lo")
```
balloob, line 59: Is this always there, even on Windows ?

aioshelly/coap.py
```diff
@@ -69,6 +84,49 @@ def socket_init():
     return sock
 
 
+class MulticastQuerier:
+    """Multicast querier management."""
+
+    def __init__(self):
+        """Initialize multicast querier thread."""
+        loop = asyncio.get_event_loop()
+        loop.run_in_executor(None, self.async_start)
```
balloob, line 93: Don't call something `async` and then run it in executor

aioshelly/coap.py
```diff
@@ -69,6 +84,49 @@ def socket_init():
...
+    @classmethod
+    def async_start(cls):
+        """Start sniffing for multicast query."""
+        filter_igmp_query = f"igmp and igmp[0] == {int(MULTICAST_QUERY_IGMPTYPE)} and (igmp[4:4] == {int(ipaddress.IPv4Address(MAIN_MULTICAST_IP))} or igmp[4:4] == 0)"
+        pkt_snd = IP(dst=MULTICAST_QUERY_GRP) / IGMP(
+            type=MULTICAST_QUERY_IGMPTYPE,
+            mrcode=100,
+            gaddr=MAIN_MULTICAST_IP,
+        )
+        while True:
```
balloob, line 104: You are not allowed to run permanent running actions in the executor. That should be an actual thread.

However, we should try to do this in asyncio. Why can't we just check in coap context if we haven't received something for X minutes and then send something again?

aioshelly/coap.py
chemelli74: I don't need any special permission to run inside HA.

aioshelly/coap.py
chemelli74: I don't have a windows system with python, but as far as I got it, "lo" is an internal name assigned by netifaces to localhost.

thecode: On windows this function returns a list of GUIDs of the interfaces which does not make sense.
In my opinion you don't need to filter lo anyhow.

aioshelly/coap.py
```diff
@@ -69,6 +86,74 @@ def socket_init():
...
+def verify_l2socket_creation_permission():
+    """Create a socket using the scapy configured l2socket.
+
+    Try to create the socket
+    to see if we have permissions
+    since sniff will do it another
+    thread so we will not be able to capture
+    any permission or bind errors.
+    """
+    conf.L2socket()
```
balloob, line 98: Do we have to close down this socket when you've checked permissions?

aioshelly/coap.py
```diff
@@ -69,6 +86,74 @@ def socket_init():
...
+
+class MulticastQuerier:
+    """Multicast querier management."""
+
+    def __init__(self):
+        """Initialize multicast querier thread."""
+        try:
+            verify_l2socket_creation_permission()
+        except (Scapy_Exception, OSError) as ex:
+            if os.geteuid() == 0:
```
balloob, line 109: Looks like you need special permissions ;-)

image

aioshelly/coap.py
```diff
@@ -69,6 +86,74 @@ def socket_init():
...
+            else:
+                _LOGGER.debug(
+                    "Cannot watch for multicast query packets without root or CAP_NET_RAW: %s",
+                    ex,
+                )
+            return
+        _LOGGER.debug("Multicast querier thread started")
+        self.stop_thread = False
+        self.thread = threading.Thread(target=self.run)
+        self.thread.start()
```
balloob, line 120: We should not start threads in the constructor. Please move this to a start method.
```suggestion

    def start(self):
        """Start the multicast querier."""
        self.thread = threading.Thread(target=self.run)
        self.thread.start()
```

aioshelly/coap.py
chemelli74: Yes, on **`venv`**. I was not thinking about manual installation of HA when I replied ;-)

aioshelly/coap.py
```diff
@@ -69,6 +86,74 @@ def socket_init():
...
+            type=MULTICAST_QUERY_IGMPTYPE,
+            mrcode=100,
+            gaddr=MAIN_MULTICAST_IP,
+        )
+        while not self.stop_thread:
+            result = sniff(
+                filter=filter_igmp_query,
+                store=1,
+                count=1,
+                timeout=MULTICAST_QUERY_TIMEOUT,
```
balloob, line 135: How long is this timeout? Because now Home Assistant shut down will wait for this timeout.

aioshelly/coap.py
```diff
@@ -69,6 +86,74 @@ def socket_init():
...
+                _LOGGER.info(
+                    "Multicast query not received in %s seconds",
+                    MULTICAST_QUERY_TIMEOUT,
+                )
+                for eth_iface in get_network_objects("iface"):
+                    send(pkt_snd, iface=eth_iface, verbose=False)
+                    _LOGGER.info("Multicast query sent for %s interface", eth_iface)
+            else:
+                _LOGGER.info(
+                    "Multicast query received from network, no action required"
```
balloob, line 148: How will this deal with things like the Shelly Motion? It's a battery device that is available on the network. Will it cause the device to wake up the whole time and acknowledge membership?

aioshelly/coap.py
```diff
@@ -79,12 +164,14 @@ def __init__(self, message_received=None):
...
         self.subscriptions = {}
         self.transport: Optional[asyncio.DatagramTransport] = None
+        self.querier = None
 
     async def initialize(self):
         """Initialize the COAP manager."""
         loop = asyncio.get_running_loop()
         self.sock = socket_init()
         await loop.create_datagram_endpoint(lambda: self, sock=self.sock)
+        self.querier = MulticastQuerier()
```
balloob, line 174:
```suggestion
        self.querier = MulticastQuerier()
        self.querier.start()
```

aioshelly/coap.py
chemelli74: Timeout is 4 minutes, but using `self_stop_thread = True` seems to work fine during shutdown and not slow down.
Any better suggestion to close the thread ?

aioshelly/coap.py
chemelli74: Multicast query are usually sent by the router, so are always present on a well working IGMP network.
In any case we don't create any traffic to those devices, as they seems not to listen to queries.

aioshelly/coap.py
chemelli74: Done.
You should review also https://github.com/home-assistant/core/blob/0382c932836ff58854fb2076a16b825928f95566/homeassistant/components/dhcp/__init__.py#L276 ;-)

aioshelly/coap.py
chemelli74: Done

aioshelly/coap.py
chemelli74: Done

chemelli74:
Seems latest commit doesn't show up:

```
To github.com:home-assistant-libs/aioshelly.git
   27de7cc..25b88ac  chemelli74-querier -> chemelli74-querier
```

aioshelly/coap.py
balloob: Mark it as a daemon and it won't prevent Python from shutting down.

aioshelly/coap.py
```diff
@@ -79,12 +170,15 @@ def __init__(self, message_received=None):
         self._message_received = message_received
         self.subscriptions = {}
         self.transport: Optional[asyncio.DatagramTransport] = None
+        self.querier = None
 
     async def initialize(self):
```
thecode, line 175: We should add a flag here (default can be true) so that we can later disable/enable querier without releasing a new version if we have problems with it.

aioshelly/coap.py
```diff
@@ -69,6 +86,80 @@ def socket_init():
...
+                    ex,
+                )
+            return
+        self.stop_thread = None
+        self.thread = None
+
+    def start(self):
+        """Start multicast querier thread."""
+        self.stop_thread = False
+        self.thread = threading.Thread(target=self.run)
```
balloob, line 124: Can you give the thread a name.

aioshelly/coap.py
thecode: @chemelli74 I think we should have a flag in `initialize` (default can be true) so that we can disable querier without releasing a new version if it makes problems.

aioshelly/coap.py
chemelli74: Done

aioshelly/coap.py
chemelli74: Done

aioshelly/coap.py
chemelli74: Moved start/stop to HA side

chemelli74: Done

aioshelly/coap.py
```diff
@@ -69,6 +87,89 @@ def socket_init():
...
+        except (Scapy_Exception, OSError) as ex:
+            if os.geteuid() == 0:
+                _LOGGER.error("Cannot watch for multicast query packets: %s", ex)
+            else:
+                _LOGGER.debug(
+                    "Cannot watch for multicast query packets without root or CAP_NET_RAW: %s",
+                    ex,
+                )
+            return
+        _lib_name = find_library("pcap")
```
balloob, line 119: Isn't this already checked in scapy?

aioshelly/coap.py
```diff
@@ -69,6 +87,89 @@ def socket_init():
...
+        self.stop_thread = None
+        self.thread = None
+
+    def start(self):
+        """Start multicast querier thread."""
+        self.stop_thread = False
+        self.thread = threading.Thread(
+            name="MulticastQuerierThread", target=self.run, daemon=True
+        )
+        _LOGGER.debug("Multicast querier thread started")
```
balloob, line 135:
```suggestion
        _LOGGER.debug("Starting multicast query thread")
```

aioshelly/coap.py
chemelli74: Unfortunately not; I added the check after other Teams member had issues on the dev container.

You think would be better to add the library to the dev container and remove the check here ?

aioshelly/coap.py
chemelli74: Done; also for stop.

chemelli74:
Team decision is to move this functionality in a custom components. At least for now.

PR #67, opened 2021-02-04
Socket binding filter by network ip

chemelli74:
On systems with more than one network interface, we need to join on all of those to be certain we receive multicast packets.

For those that have many network interfaces, this approach unfortunately consumes too much network resources resulting in 
`OSError: [Errno 105] No buffer space available`

Some users already found a workarround: increase the OS max multicast group join parameter.
Thus confirming that limiting the binding will solve the issue.

This PR will filter all those interfaces that doesn't have a suitable IP for networking operations.

### DIFF
--- a/aioshelly/coap.py
+++ b/aioshelly/coap.py
@@ -7,11 +7,16 @@ import struct
 from typing import Optional, cast
 
 import netifaces
+from netaddr import IPAddress, IPNetwork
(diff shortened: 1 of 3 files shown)


chemelli74:
HA is currently supporting only 1 network interface, please review your routing accordly

Issue #72, opened 2021-02-13
Logger: aioshelly.coap


FCKotze:
Logger: aioshelly.coap
Source: /usr/local/lib/python3.8/site-packages/aioshelly/coap.py:29
First occurred: 11:47:03 AM (1 occurrences)
Last logged: 11:47:03 AM

CoAP message of type 69 from host 192.168.86.234 is not a valid JSON format: b"XE'\x1e\t\x0c\xdf=q\x11\xe1\xc2\xb3oic\x02ad\x11<\xe1\xfe\xdb\xc0\xff\xbfcnon\x1a\x94\xa2L\xe6cttl\x18<ctrg\xe0brtioic.wk.ad\xff"

I have this error pop up in my Home Assistant Logs a few times a day, when I check which device it related to based on the IP Address listed its my Samsung Robot Vacuum. I decided to double check to make sure its the vacuum causing this error by assigning it a different IP address and yes the new IP address i assigned then gets listed on the error. I have no idea why my Samsung Robot Vacuum would give a Shelly Integration Error. Nothing stops working and no entities become unavailable, yet this error message would appear several times a day. Hope you would be able to shed some light on this issue for m. Thanks

thecode:
Thanks @FCKotze, I will issue a fix to filter only Shelly IPs in the next days

FCKotze:
Thanks @thecode , much appreciated!

balloob:
But then we can't detect new Shelly devices anymore with our coap listener. Maybe only print as warning if the IP is not known to be Shelly?

thecode:
Currently we do not use CoAP to detect new devices, but indeed we had a plan for it.
What we do now is we parse the message here:
https://github.com/home-assistant-libs/aioshelly/blob/1d25c9030870b4640ddace243664adfcd82013ee/aioshelly/coap.py#L88
And if the IP is not in the list we drop the message here:
https://github.com/home-assistant-libs/aioshelly/blob/1d25c9030870b4640ddace243664adfcd82013ee/aioshelly/coap.py#L97

The reason I did not submit a fix yet is that is that indeed it will remove the option to detect devices using CoAP. 
The problem with using port 5683 is that it is a standard CoAP port, and although now in home usage we only see it with Shelly, it maybe be used by other devices in the future.

balloob:
I think in that case, we should not log it as a warning but debug and silently discard the message.

chemelli74:
[...]

PR #75, opened 2021-02-22
Fix coap logging

thecode:
Change CoAP message parsing log level from error to debug to avoid logging errors for devices which are not Shelly devices and sends CoAP messages.
This PR fixes issue: https://github.com/home-assistant-libs/aioshelly/issues/72

### DIFF
--- a/aioshelly/coap.py
+++ b/aioshelly/coap.py
@@ -26,7 +26,7 @@ class CoapMessage:
             try:
                 self.payload = json.loads(payload.rsplit(b"\xff", 1)[1].decode())
             except (json.decoder.JSONDecodeError, IndexError):
-                _LOGGER.error(
+                _LOGGER.debug(
                     "CoAP message of type %s from host %s is not a valid JSON format: %s",
                     self.code,
                     self.ip,
--- a/aioshelly/coap.py
+++ b/aioshelly/coap.py
@@ -9,6 +9,14 @@ from typing import Optional, cast
 _LOGGER = logging.getLogger(__name__)
 
 
+class CoapError(Exception):
+    """Base class for COAP errors."""
+
+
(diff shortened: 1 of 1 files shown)


aioshelly/coap.py
```diff
@@ -26,7 +26,7 @@ def __init__(self, sender_addr, payload: bytes):
             try:
                 self.payload = json.loads(payload.rsplit(b"\xff", 1)[1].decode())
             except (json.decoder.JSONDecodeError, IndexError):
-                _LOGGER.error(
+                _LOGGER.debug(
```
balloob, line 29: It's weird that we set payload to None instead of raising this error if we receive invalid JSON. Now we need to guard for payload = None later to detect if this went wrong. Shouldn't we just throw an `InvalidMessage` error?

thecode: We indeed guard for it right after parsing:
https://github.com/home-assistant-libs/aioshelly/blob/4ea230e5f88e8dbaac6ddb3d2e601345644982e8/aioshelly/coap.py#L91

I agree that raising an `InvalidMessage` is a better approach but than we need to rewrite this function to change this also:
https://github.com/home-assistant-libs/aioshelly/blob/4ea230e5f88e8dbaac6ddb3d2e601345644982e8/aioshelly/coap.py#L37

Do you think it is better to rewrite it and catch all errors here:
https://github.com/home-assistant-libs/aioshelly/blob/4ea230e5f88e8dbaac6ddb3d2e601345644982e8/aioshelly/coap.py#L88

So that we can raise errors if the IP is in the list or silently log it if it is not?

balloob: Yeah that sounds better.

balloob (approved):

PR #86, opened 2021-03-28
Shelly 4Pro and Shelly Sense are unsupported

chemelli74:
Force Shelly4Pro and Shelly Sense unsupported firmware message based on device type and not on firmware release date.

Those devices will never get CoAP v2, but still get some firmware update


bieniu (approved):
LGTM

thecode (approved):

PR #87, opened 2021-04-01
Fix device initialize

thecode:
Make sure ` self._initialized` is cleared upon exceptions during initialize.
Catch errors when `self.initialize` is called upon CoAP event and log as warning.

### DIFF
--- a/aioshelly/__init__.py
+++ b/aioshelly/__init__.py
@@ -1,14 +1,18 @@
 """Shelly CoAP library."""
 import asyncio
-import ipaddress
-import re
 from dataclasses import dataclass
+import ipaddress
+import logging
+import re
 from socket import gethostbyname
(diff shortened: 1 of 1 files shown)


aioshelly/__init__.py
```diff
@@ -229,11 +237,21 @@ def shutdown(self):
         self._update_listener = None
         self._unsub_listening()
 
+    async def _async_init(self):
+        """Async init upon CoAP message event."""
+        try:
+            async with async_timeout.timeout(10):
```
chemelli74, line 243: I would suggest using a general constant so that if we need it in other places in future we are sure to use the same value

balloob (approved):

balloob:
If you do a new release for HA, DM me on Discord and I'll tag it for the beta.

PR #90, opened 2021-04-05
Add trigger_ota_update device method

mib1185:
This PR will add a new device method `trigger_ota_update` which allows to trigger an ota update according to https://shelly-api-docs.shelly.cloud/#ota.

#### without parameters
Run firmware update from default URL  (`new_version`)
#### with given `url` parameter
Run firmware update from specified URL
#### with given `beta` parameter
Run firmware update from beta URL (if available)
#### with both given `beta` and `url` parameter
Still run firmware update from specified URL (_IMHO because url is more specific than beta_)

### DIFF
--- a/aioshelly/__init__.py
+++ b/aioshelly/__init__.py
@@ -346,6 +346,17 @@ class Device:
         """Change device mode color/white."""
         return await self.http_request("get", "settings", {"mode": mode})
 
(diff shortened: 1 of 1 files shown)


thecode (approved):
Thanks @mib1185

PR #102, opened 2021-05-23
Add docker files

chemelli74:
Add docker files to allow a easy "`Remote-Containers: Clone Repository in Container Volume`" from VScode


thecode (approved):
Great job, will make development easier and faster

PR #127, opened 2021-09-07
Enable strict typing

chemelli74:


balloob:
Should we wait for #128 ?

thecode:
> Should we wait for #128 ?

Yes we wait for #128

aioshelly/wsrpc.py
```diff
@@ -231,7 +231,7 @@ async def call(
 
         if "result" in resp:
             _LOGGER.debug("%s(%s) -> %s", call.method, call.params, resp["result"])
-            return resp["result"]
+            return dict(resp["result"])
```
bieniu, line 234: I think we should use `cast()` instead of `dict()`.

thecode (approved):
LGTM

aioshelly/wsrpc.py
chemelli74: Fixed

thecode (approved):

bieniu (approved):
LGTM 👍

PR #128, opened 2021-09-07
Add Shelly Gen2 Support

thecode:
## Overview
Shelly Gen2 devices are based on RPC over Web socket and are not backward compatible with Gen1 (CoAP based) devices.
This is a refactor of the `aioshelly` library and has breaking changes. Library will support both Gen1 & Gen2 devices.

More details about Gen2 devices API can be found here: https://shelly-api-docs.shelly.cloud/gen2/
## Files content
Library is built from the following files:
- `__init__.py` - holds a Device class which creates either Gen1 or Gen2 device
- `coap.py` - handles low level CoAP communication with Gen1 devices (unchanged)
- `block_device.py` - Block based Gen1 device class (previously located under `__init__.py`)
- `wsrpc.py` - handles low level Web socket RPC communication with Gen2 devices (new)
- `rpc_device.py` - RPC based Gen2 device class (new)
- `common` - common functions for both Gen1 & Gen2 devices (previously located under `__init__.py`)
- `const.py` - constants and model names for both Gen1 & Gen2 devices (previously located under `__init__.py`)
- `exceptions.py`- exceptions for both Gen1 & Gen2 devices (previously located under `__init__.py`)

## Breaking changes
Due to refactoring for supporting Gen2 devices the main device class is now a new class and most content for Block based devices is moved to `block_device`, in most cases this results by imports using Gen1 block device changed, major changes:
- `from aioshelly import Block` is changed to `from aioshelly.block_device import Block`
- Constants should be imported from `aioshelly.const` instead from `aioshelly`
- Exceptions should be imported from `aioshelly.exceptions` instead from `aioshelly`
- to support Gen2 devices, `device.shutdown()` is now `async` and should be awaited
- `request_s` which was for internal use is removed from `device.initialize`
- ` ConnectionOptions` and `get_info` are now at `common.py` instead of `__init__.py`

## Testing
`Device.create` has a new parameter `gen` which is set to `1` by default, if set to `2` it will create a Gen2 device class.
`example.py` updated to support Gen2 devices, testing Gen2 devices can be done by adding `--gen2` or `-g2` to the command line parameters.


aioshelly/__init__.py
```diff
@@ -1,187 +1,30 @@
...
-        self._unsub_listening = coap_context.subscribe_updates(
-            options.ip_address, self._coap_message_received
-        )
-        self._update_listener = None
-        self._coap_response_events: dict = {}
-        self._initialized = False
-        self._initializing = False
-
     @classmethod
     async def create(
```
balloob, line 20: This makes no sense to keep this on this device. Instead, I suggest we split the function and move it to `BlockDevice` and `RpcDevice`, also because the parameters are different.

aioshelly/__init__.py
```diff
@@ -1,187 +1,30 @@
...
         cls,
         aiohttp_session: aiohttp.ClientSession,
         coap_context: COAP,
         ip_or_options: Union[str, ConnectionOptions],
         initialize: bool = True,
-    ):
+        gen=1,
+    ):  # pylint: disable=too-many-arguments
         """Device creation."""
         if isinstance(ip_or_options, str):
```
balloob, line 29: This part can be moved into a util function to process ip_or_options.

aioshelly/const.py
```diff
@@ -0,0 +1,56 @@
...
+    "SHUNI-1": "Shelly UNI",
+    "SHVIN-1": "Shelly Vintage",
+    "SHWT-1": "Shelly Flood",
+    # Gen2 RPC based models
+    "SNSW-001P16EU": "Shelly Plus 1PM",
+    "SNSW-001X16EU": "Shelly Plus 1",
+    "SPSW-004PE16EU": "Shelly Pro 4PM",
+}
+
+DEVICE_TIMEOUT_SEC = 10
```
balloob, line 51: Is this time out for HTTP, RPC or COAP?

aioshelly/rpc_device.py
```diff
@@ -0,0 +1,202 @@
...
+import aiohttp
+
+from .common import ConnectionOptions, get_info
+from .exceptions import AuthRequired, NotInitialized
+from .wsrpc import WsRPC
+
+
+def mergedicts(dict1, dict2):
+    """Deep dicts merge."""
+    for k in set(dict1.keys()).union(dict2.keys()):
```
balloob, line 15:
```suggestion
    for k in set(dict1).union(dict2):
```

aioshelly/rpc_device.py
```diff
@@ -0,0 +1,202 @@
...
+        if k in dict1 and k in dict2:
+            if isinstance(dict1[k], dict) and isinstance(dict2[k], dict):
+                yield (k, dict(mergedicts(dict1[k], dict2[k])))
```
balloob, line 18: why is this yielding? Now you're always going to call it with `dict` wrapping the generator ?

aioshelly/rpc_device.py
```diff
@@ -0,0 +1,202 @@
...
+            else:
+                yield (k, dict2[k])
+        elif k in dict1:
+            yield (k, dict1[k])
+        else:
+            yield (k, dict2[k])
```
balloob, line 24: Seems that this is slightly simpler?

```suggestion
    result = dict(dict1)
    result.update(dict2)
    for key, value in result.items():
        if isinstance(value, dict) and isinstance(dict1.get(key), dict):
            result[key] = mergedicts2(dict1[key], value)
    return result
```

aioshelly/rpc_device.py
```diff
@@ -0,0 +1,202 @@
...
+            self._update_listener(self)
+
+    @property
+    def ip_address(self):
+        """Device ip address."""
+        return self.options.ip_address
+
+    async def initialize(self):
+        """Device initialization."""
+        self._initializing = True
```
balloob, line 79: Maybe helpful guard:
```suggestion
		if self._initializing:
		    raise RuntimeError("Already initializing")

        self._initializing = True
```

aioshelly/rpc_device.py
```diff
@@ -0,0 +1,202 @@
...
+        self._initialized = False
+        try:
+            await self._wsrpc.connect(self.aiohttp_session)
+            self.shelly = await get_info(self.aiohttp_session, self.options.ip_address)
+
+            if self.options.auth or not self.shelly["auth_en"]:
+                await self.update_device_info()
+                await self.update_config()
+                await self.update_status()
```
balloob, line 88: Should we do it in parallel? We used to not do that because of COAP response issues, but we don't have that here? 
```suggestion
                await asyncio.gather(
                    self.update_device_info(),
                    self.update_config(),
                    self.update_status(),
                )
```

aioshelly/rpc_device.py
```diff
@@ -0,0 +1,202 @@
...
+        self._device_info = await self._wsrpc.call("Shelly.GetDeviceInfo")
+
+    async def update_config(self):
+        """Get device config from 'Shelly.GetConfig'."""
+        self._config = await self._wsrpc.call("Shelly.GetConfig")
+
+    @property
+    def initialized(self):
+        """Device initialized."""
+        return self._initialized
```
balloob, line 121: Instead of a property you could just store it as `self.initalized` 🤷

aioshelly/rpc_device.py
```diff
@@ -0,0 +1,202 @@
...
+
+    @property
+    def requires_auth(self):
+        """Device check for authentication."""
+        return self.shelly["auth_en"]
+
+    async def set_state(self, method, params):
```
balloob, line 128: Why is this called `set_state` and not call_rpc or similar?

aioshelly/wsrpc.py
```diff
@@ -0,0 +1,194 @@
...
+
+    @property
+    def _next_id(self):
+        self._call_id += 1
+        return self._call_id
+
+    async def connect(self, aiohttp_session):
+        """Connect to device."""
+        if self.connected:
+            _LOGGER.debug("%s already connected", self._ip_address)
```
balloob, line 72: I would raise instead. We should not end up calling connect twice

aioshelly/wsrpc.py
```diff
@@ -0,0 +1,194 @@
...
+            return
+
+        _LOGGER.debug("Trying to connect to device at %s", self._ip_address)
+        try:
+            self._websocket = await aiohttp_session.ws_connect(
+                f"http://{self._ip_address}/rpc", heartbeat=WS_HEARTBEAT_SEC
```
balloob, line 78: In aiohttp the heartbeat is broken. It will send a ping but if it times out it won't cancel any open receive command. If the connection is closed it will print a stack trace that's not tracked. 

Instead I would follow this approach:
https://github.com/NabuCasa/hass-nabucasa/blob/master/hass_nabucasa/iot_base.py#L180-L185

aioshelly/wsrpc.py
```diff
@@ -0,0 +1,194 @@
...
+    async def _rx_msgs(self):
+        async for msg in self._websocket:
+            _LOGGER.debug("Received Message: Type: %s,  Data: %s", msg.type, msg.data)
+
+            try:
+                frame = msg.json()
+            except ValueError as err:
+                raise InvalidMessage("Received invalid JSON.") from err
+            else:
+                await self._handle_frame(frame)
```
balloob, line 155: This should not be async. By making it async you now have to wait for any callback to finish before you process the next message, instead of being able to process the 2nd message while the 1st message is processed in the background etc. Preferably you keep `_handle_frame` fully non-async, and let the listeners of the future decide if they want to be async.

aioshelly/wsrpc.py
```diff
@@ -0,0 +1,194 @@
...
+
+        error = str(self._websocket.exception()) if self._websocket else "Disconnected"
+        _LOGGER.debug("Websocket error: %s", error)
+        await self._on_notification("WebSocketClosed", {"error": error})
```
balloob, line 159: Since these are your own notifications, let's define a constant for `WebSocketClosed`?

aioshelly/wsrpc.py
```diff
@@ -0,0 +1,194 @@
...
+
+    @property
+    def connected(self) -> bool:
+        """Return if we're currently connected."""
+        return self._websocket is not None and not self._websocket.closed
+
+    async def call(self, method, params=None, timeout=10):
+        """Websocket RPC call."""
+        call = RPCCall(self._next_id, method, params)
+        call.resolve = asyncio.Future()
```
balloob, line 169: IF these 3 things are always set, why not put them in the constructor / use a factory to generate it as part of constructing an instance ?

aioshelly/wsrpc.py
```diff
@@ -0,0 +1,194 @@
...
+        call.src = self._src
+        call.dst = self._dst
+
+        self._calls[call.call_id] = call
+        await self._websocket.send_json(call.request_frame)
+        call.sent_at = datetime.utcnow()
+
+        try:
+            resp = await asyncio.wait_for(call.resolve, timeout)
```
balloob, line 178: We generally use `async_timeout`, it's a nicer syntax and is already installed as part of aiohttp;

```python
import async_timeout

async with async_timeout.timeout(timeout):
    await call.resolve
```

aioshelly/wsrpc.py
```diff
@@ -0,0 +1,194 @@
...
+        else:
+            _LOGGER.warning("Invalid frame: %s", frame)
+
+    async def _rx_msgs(self):
+        async for msg in self._websocket:
+            _LOGGER.debug("Received Message: Type: %s,  Data: %s", msg.type, msg.data)
+
+            try:
+                frame = msg.json()
+            except ValueError as err:
```
balloob, line 152: Is it just `ValueError`, even if it's a CLOSING message for example?

aioshelly/wsrpc.py
```diff
@@ -0,0 +1,194 @@
...
+            }
+        )
+
+    async def _handle_frame(self, frame):
+        if peer_src := frame.get("src", None):
+            if self._dst is not None and peer_src != self._dst:
+                _LOGGER.warning("Remote src changed: %s -> %s", self._dst, peer_src)
+            self._dst = peer_src
+
+        frame_id = frame.get("id", None)
```
balloob, line 119: `None` is the default (assuming `frame` is a dict)

aioshelly/wsrpc.py
```diff
@@ -0,0 +1,194 @@
...
+        method = frame.get("method", None)
+
+        if method:
```
balloob, line 122: You could walrus this :D

aioshelly/wsrpc.py
```diff
@@ -0,0 +1,194 @@
...
+        self._call_id = 1
+        self._src = f"aios-{id(self)}"
+        self._dst = None
+
+    @property
+    def _next_id(self):
+        self._call_id += 1
+        return self._call_id
+
+    async def connect(self, aiohttp_session):
```
balloob, line 69: If you want, you could simplify the logic of this class if you just make `connect` do all the work and not create a `self._rx_task`. That way the disconnect event is built-in -> the method finishes. It's what we do for ZJS https://github.com/home-assistant-libs/zwave-js-server-python/blob/master/zwave_js_server/client.py#L162  

You could even make it go one up and make it an async iterator that iterates over events and then in the end just finishes when connection done. We do that in aiohue: https://github.com/home-assistant-libs/aiohue/blob/master/aiohue/clip.py#L11 and consume it [like this](https://github.com/home-assistant-libs/aiohue/blob/master/aiohue/bridge.py#L162)

aioshelly/wsrpc.py
```diff
@@ -0,0 +1,194 @@
...
+            raise CannotConnect(f"Error connecting to {self._ip_address}") from err
+
+        self._rx_task = asyncio.create_task(self._rx_msgs())
+
+        _LOGGER.info("Connected to %s", self._ip_address)
+
+    async def disconnect(self):
+        """Disconnect all sessions."""
+        if self._websocket is None:
+            return
```
balloob, line 93: Avoid silent error handling. Raise because the code calling this is not keeping proper track of their objects.

```python
if not self.connected:
```

aioshelly/wsrpc.py
```diff
@@ -0,0 +1,194 @@
...
+        async for msg in self._websocket:
+            _LOGGER.debug("Received Message: Type: %s,  Data: %s", msg.type, msg.data)
+
+            try:
+                frame = msg.json()
+            except ValueError as err:
+                raise InvalidMessage("Received invalid JSON.") from err
+            else:
+                await self._handle_frame(frame)
+
```
balloob, line 156: When you finish the `async for`, you should cancel all open calls. This should not happen in the `disconnect()` method, because a disconnect can also be caused by loss of internet/device rebooting.

aioshelly/wsrpc.py
```diff
@@ -0,0 +1,194 @@
...
+    async def call(self, method, params=None, timeout=10):
+        """Websocket RPC call."""
+        call = RPCCall(self._next_id, method, params)
+        call.resolve = asyncio.Future()
+        call.src = self._src
+        call.dst = self._dst
+
+        self._calls[call.call_id] = call
+        await self._websocket.send_json(call.request_frame)
+        call.sent_at = datetime.utcnow()
```
balloob, line 175: this data is never used ?

aioshelly/rpc_device.py
thecode: Yes that will work

aioshelly/rpc_device.py
thecode: Changed to `call_rpc`

aioshelly/wsrpc.py
thecode: No, was already thinking to remove it, removed

aioshelly/const.py
thecode: General device timeout, common to HTTP/RPC & CoAP, added comment for that

aioshelly/wsrpc.py
thecode: No, in case of `MsgType.CLOSE`, `WSMsgType.CLOSING`, `WSMsgType.CLOSED` we should just stop processing messages

aioshelly/const.py
thecode: Actually only used in Block device init, renamed to `BLOCK_DEVICE_INIT_TIMEOUT`

aioshelly/wsrpc.py
thecode: I think this changes too much the current implementation and the benefit from this is small so let's leave that to a future change if we make it

balloob (approved):

PR #132, opened 2021-09-10
Strict typing part III: wsrpc.py

chemelli74:


aioshelly/wsrpc.py
```diff
@@ -63,22 +65,22 @@ def request_frame(self):
...
 
-    def __init__(self, ip_address: str, on_notification):
+    def __init__(self, ip_address: str, on_notification: Callable) -> None:
         """Initialize WsRPC class."""
         self._ip_address = ip_address
         self._on_notification = on_notification
-        self._rx_task = None
+        self._rx_task: tasks.Task[None] | None = None
         self._client: ClientWebSocketResponse | None = None
         self._calls: Dict[str, RPCCall] = {}
```
thecode, line 74:
```suggestion
        self._calls: Dict[int, RPCCall] = {}
```

aioshelly/wsrpc.py
```diff
@@ -202,10 +210,16 @@ def connected(self) -> bool:
...
         return self._client is not None and not self._client.closed
 
-    async def call(self, method, params=None, timeout=10):
+    async def call(
+        self, method: str, params: dict[str, Any] | None = None, timeout: int = 10
+    ) -> Any:
         """Websocket RPC call."""
         call = RPCCall(self._next_id, method, params, self._route)
-        self._calls[call.call_id] = call
+        self._calls[str(call.call_id)] = call
```
thecode, line 218:
```suggestion
        self._calls[call.call_id] = call
```

thecode (changes requested):
Let's change the dict key to int so we don't need to cast later

aioshelly/wsrpc.py
```diff
@@ -108,7 +110,10 @@ async def disconnect(self):
 
         self._rx_task = None
 
-    async def _handle_call(self, frame_id):
+    async def _handle_call(self, frame_id: str) -> None:
+        if not self._client:
+            return
```
thecode, line 115: This can't happen so it is better to assert here and not just return

```suggestion
        assert self._client
```

aioshelly/wsrpc.py
```diff
@@ -151,7 +156,10 @@ def _handle_frame(self, frame):
         else:
             _LOGGER.warning("Invalid frame: %s", frame)
 
-    async def _rx_msgs(self):
+    async def _rx_msgs(self) -> None:
+        if not self._client:
+            return
```
thecode, line 161: This can't happen so it is better to assert here and not just return
```suggestion
        assert self._client
```

aioshelly/wsrpc.py
```diff
@@ -202,10 +210,16 @@ def connected(self) -> bool:
         """Return if we're currently connected."""
         return self._client is not None and not self._client.closed
 
-    async def call(self, method, params=None, timeout=10):
+    async def call(
+        self, method: str, params: dict[str, Any] | None = None, timeout: int = 10
+    ) -> Any:
         """Websocket RPC call."""
         call = RPCCall(self._next_id, method, params, self._route)
```
thecode, line 217: Let's make sure we raise an error since this is called from outside of the class

```suggestion
        if self._client is None:
            raise RuntimeError("Not connected")

        call = RPCCall(self._next_id, method, params, self._route)
```

aioshelly/wsrpc.py
```diff
@@ -202,10 +210,16 @@ def connected(self) -> bool:
...
         self._calls[call.call_id] = call
+
+        if not self._client:
+            return
+
```
thecode, line 222: This must happen before the `self._calls` is assigned and added a comment to raise error so it is not needed
```suggestion

```

thecode (changes requested):

aioshelly/wsrpc.py
```diff
@@ -202,10 +210,16 @@ def connected(self) -> bool:
         """Return if we're currently connected."""
         return self._client is not None and not self._client.closed
 
-    async def call(self, method, params=None, timeout=10):
+    async def call(
+        self, method: str, params: dict[str, Any] | None = None, timeout: int = 10
+    ) -> Any:
```
thecode, line 215: No need to return `Any` if we add raise on error
```suggestion
    ) -> dict[str, Any]:
```

thecode:
Also add cast here:
https://github.com/home-assistant-libs/aioshelly/blob/aa792af119a4ca0fa35c484fd6ffe88be2de8b54/aioshelly/wsrpc.py#L237
```python
return cast(dict, resp["result"])
```

thecode (approved):
Looks good, Thanks

PR #133, opened 2021-09-10
Strict typing part IV: rpc_device.py

chemelli74:


aioshelly/rpc_device.py
```diff
@@ -2,18 +2,21 @@
...
+from aiohttp.client import ClientSession
 
 from .common import ConnectionOptions, IpOrOptionsType, get_info, process_ip_or_options
-from .exceptions import AuthRequired, NotInitialized, WrongShellyGen
+from .exceptions import AuthRequired, NotInitialized, ShellyError, WrongShellyGen
 from .wsrpc import WsRPC
 
 
-def mergedicts(dict1, dict2):
+def mergedicts(dict1: dict, dict2: dict | None) -> dict:
```
thecode, line 15: No need to allow `None` here for `dict2`
```suggestion
def mergedicts(dict1: dict, dict2: dict) -> dict:
```

aioshelly/rpc_device.py
```diff
@@ -2,18 +2,21 @@
...
     """Deep dicts merge."""
     result = dict(dict1)
+    if dict2 is None:
+        return result
```
thecode, line 19: `dict2` can't be `None` (see typing change above)
```suggestion
```

aioshelly/rpc_device.py
```diff
@@ -58,22 +61,26 @@ async def create(
...
-    def _on_notification(self, method, params=None):
+    def _on_notification(
+        self, method: str, params: dict[str, Any] | None = None
+    ) -> None:
         if params is not None:
             if method == "NotifyStatus":
-                self._status = dict(mergedicts(self._status, params))
+                self._status = (
+                    dict(mergedicts(self._status, params)) if self._status else None
+                )
```
thecode, line 71: We can't reach here if `self._status` is `None` so it is better to assert it
```suggestion
                assert self._status
                self._status = dict(mergedicts(self._status, params))
```

aioshelly/rpc_device.py
```diff
@@ -97,41 +104,41 @@ async def initialize(self):
...
         """Device check for authentication."""
-        if "auth_en" not in self.shelly:
+        if not self.shelly or "auth_en" not in self.shelly:
             raise WrongShellyGen
 
-        return self.shelly["auth_en"]
+        return bool(self.shelly["auth_en"])
 
-    async def call_rpc(self, method, params):
+    async def call_rpc(self, method: str, params: Any) -> dict[str, Any]:
```
thecode, line 136:
```suggestion
    async def call_rpc(self, method: str, params: dict[str, Any] | None) -> dict[str, Any]:
```

aioshelly/rpc_device.py
```diff
@@ -172,32 +179,38 @@ def config(self):
...
+    def firmware_version(self) -> str:
         """Device firmware version."""
+        if self.shelly is None:
+            raise ShellyError
+
         if not self.initialized:
             raise NotInitialized
 
-        return self.shelly["fw_id"]
+        return str(self.shelly["fw_id"])
```
thecode, line 195:
```suggestion
        if self.shelly is None:
            raise AuthRequired

        return cast(str, self.shelly["fw_id"])
```

aioshelly/rpc_device.py
```diff
@@ -172,32 +179,38 @@ def config(self):
...
 
-        return self.shelly["model"]
+        return str(self.shelly["model"])
 
     @property
-    def hostname(self):
+    def hostname(self) -> str:
         """Device hostname."""
-        return self.device_info["id"]
+        return str(self.device_info["id"])
```
thecode, line 211:
```suggestion
        return cast(str, self.device_info["id"])
```

aioshelly/rpc_device.py
```diff
@@ -172,32 +179,38 @@ def config(self):
...
+    def model(self) -> str:
         """Device model."""
+        if self.shelly is None:
+            raise ShellyError
+
         if not self.initialized:
             raise NotInitialized
 
-        return self.shelly["model"]
+        return str(self.shelly["model"])
```
thecode, line 206:
```suggestion
        if self.shelly is None:
            raise AuthRequired

        return cast(str, self.shelly["model"])
```

aioshelly/rpc_device.py
```diff
@@ -2,18 +2,21 @@
...
 import asyncio
-from typing import Any, Dict
+from typing import Any, Callable
 
 import aiohttp
+from aiohttp.client import ClientSession
 
 from .common import ConnectionOptions, IpOrOptionsType, get_info, process_ip_or_options
-from .exceptions import AuthRequired, NotInitialized, WrongShellyGen
+from .exceptions import AuthRequired, NotInitialized, ShellyError, WrongShellyGen
```
thecode, line 11: It is better to use the `AuthRequired` as we use in other places and not generic error
```suggestion
from .exceptions import AuthRequired, NotInitialized, WrongShellyGen
```

aioshelly/rpc_device.py
```diff
@@ -172,32 +179,38 @@ def config(self):
...
-        return self.shelly["fw_id"]
+        return str(self.shelly["fw_id"])
 
     @property
-    def model(self):
+    def model(self) -> str:
         """Device model."""
+        if self.shelly is None:
+            raise ShellyError
+
```
thecode, line 202: Let's raise `AuthRequired` after we check if device is initialized same as we do for `device_info` and `config`
```suggestion
```

aioshelly/rpc_device.py
```diff
@@ -172,32 +179,38 @@ def config(self):
...
         """Device generation: GEN2 - RPC."""
         return 2
 
     @property
-    def firmware_version(self):
+    def firmware_version(self) -> str:
         """Device firmware version."""
+        if self.shelly is None:
+            raise ShellyError
+
```
thecode, line 191: Let's raise `AuthRequired` after we check if device is initialized same as we do for `device_info` and `config`
```suggestion
```

thecode (changes requested):

aioshelly/rpc_device.py
chemelli74: IMHO, If Shelly is missing is not a auth error, but a generic error

aioshelly/rpc_device.py
thecode: Offline discussion - `self.shelly` can't be `None`, not related to auth

thecode (approved):
Looks perfect! Thanks

PR #142, opened 2021-09-29
Add heartbeat to WsRPC

thecode:
Add heartbeat to WsRPC to improve detection of disconnected device

### DIFF
--- a/aioshelly/wsrpc.py
+++ b/aioshelly/wsrpc.py
@@ -12,7 +12,7 @@ import aiohttp
 import async_timeout
 from aiohttp import ClientWebSocketResponse, WSMsgType, client_exceptions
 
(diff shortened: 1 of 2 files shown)


aioshelly/wsrpc.py
```diff
@@ -88,7 +88,7 @@ async def connect(self, aiohttp_session: aiohttp.ClientSession) -> None:
         _LOGGER.debug("Trying to connect to device at %s", self._ip_address)
         try:
             self._client = await aiohttp_session.ws_connect(
-                f"http://{self._ip_address}/rpc"
+                f"http://{self._ip_address}/rpc", heartbeat=WS_HEARTBEAT
```
balloob, line 91: If you do this, you can drop the manual added heartbeat with timeout

balloob: 162-164 + 185

thecode: Removed

PR #161, opened 2021-11-05
Add trigger reboot call

mib1185:
this will add `trigger_reboot()`  to block and rpc devices.

### DIFF
--- a/aioshelly/block_device.py
+++ b/aioshelly/block_device.py
@@ -245,6 +245,10 @@ class BlockDevice:
 
         return await self.http_request("get", "ota", params=params)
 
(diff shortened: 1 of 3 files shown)


mib1185:
should we disable `too-many-public-methods` checks in pylint at all, or only increase from 20 (default) to maybe 25?

chemelli74:
> should we disable `too-many-public-methods` checks in pylint at all, or only increase from 20 (default) to maybe 25?

didn't check, but maybe something can be moved to private.

Simone

thecode (approved):

chemelli74 (approved):

PR #163, opened 2021-11-08
Example.py: allow CoAP port customization and debug logging level

chemelli74:
2 new parameters for example.py:

- "`--socket_port`" : allows CoAP port customization.
- "`--debug`": set debug logging level.

### DIFF
--- a/example.py
+++ b/example.py
@@ -6,8 +6,11 @@ import argparse
 import asyncio
 import json
 import logging
+import signal
+import sys
 import traceback
 from datetime import datetime
+from types import FrameType
 from typing import Any, Tuple, cast
 
 import aiohttp
@@ -24,6 +27,14 @@ from aioshelly.rpc_device import RpcDevice
 async def get_coap_context(port: int) -> COAP:
     """Create CoAP context"""
     context = COAP()
+
+    def handle_sigint(_exit_code: int, _frame: FrameType) -> None:
+        """Handle Keyboard signal interrupt (ctrl-c)."""
+        context.close()
+        sys.exit()
+
+    signal.signal(signal.SIGINT, handle_sigint)
+
     await context.initialize(port)
     return context
 
@@ -271,7 +282,4 @@ async def main() -> None:
 
 
 if __name__ == "__main__":
-    try:
-        asyncio.run(main())
-    except KeyboardInterrupt:
-        pass
+    asyncio.run(main())
--- a/example.py
+++ b/example.py
@@ -65,14 +65,10 @@ async def test_single(
     """Test single device."""
     async with aiohttp.ClientSession() as aiohttp_session:
         try:
+            coap_context = await get_coap_context(port)
             async with async_timeout.timeout(timeout):
                 device = await create_device(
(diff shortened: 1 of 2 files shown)


thecode:
I find it very odd that we know have two places for setting the port, one in the constructor and one in the `initialize` function.

The problem was that from the first we should have port the port in the constructor.

Consider this code:
```python
coap = COAP(1234)
coap.initialize(1111)
```
We have now speficied two different ports and the port from the constructor will override the port  set by `initialize`

I think we should consider one of the following:
1. Move the port to the constructor and remove it from `initialize`, clean solution,  **breaking change**
2. Modify the example to pass the port via `initialize` (without using asynchronous context manager) - simple change

example.py
```diff
@@ -216,6 +225,9 @@ def get_arguments() -> Tuple[argparse.ArgumentParser, argparse.Namespace]:
     parser.add_argument(
         "--gen2", "-g2", action="store_true", help="Force Gen 2 (RPC) device"
     )
+    parser.add_argument(
+        "--debugger", "-deb", action="store_true", help="Enable debug level for logging"
```
thecode, line 229: This not really enable a debugger, just set a debug log, so it would be better to call it `--debug`

chemelli74:
> I find it very odd that we know have two places for setting the port, one in the constructor and one in the `initialize` function.

Agree

> I think we should consider one of the following:
> 
> 1. Move the port to the constructor and remove it from `initialize`, clean solution,  **breaking change**
> 2. Modify the example to pass the port via `initialize` (without using asynchronous context manager) - simple change

I prefer the second option, so I implemented it ;-)

Simone

example.py
```diff
@@ -20,6 +21,13 @@
 from aioshelly.rpc_device import RpcDevice
 
 
+async def get_coap_context(port: int | None = None) -> COAP:
```
thecode, line 24:
```suggestion
async def get_coap_context(port: int) -> COAP:
```

example.py
```diff
@@ -75,19 +89,24 @@ async def test_single(
...
+async def test_devices(init: bool, timeout: float, port: int, gen: int | None) -> None:
     """Test multiple devices."""
     device_options = []
     with open("devices.json") as fp:
         for line in fp:
             device_options.append(ConnectionOptions(**json.loads(line)))
 
-    async with aiohttp.ClientSession() as aiohttp_session, COAP() as coap_context:
+    async with aiohttp.ClientSession() as aiohttp_session:
         results = await asyncio.gather(
```
thecode, line 100:
```suggestion
        coap_context = await get_coap_context(port)
        results = await asyncio.gather(
```

example.py
```diff
@@ -75,19 +89,24 @@ async def test_single(
...
             *[
                 asyncio.wait_for(
                     connect_and_print_device(
-                        aiohttp_session, coap_context, options, init, timeout, gen
+                        aiohttp_session,
+                        await get_coap_context(port),
```
thecode, line 105:
```suggestion
                        coap_context,
```

thecode (approved):

PR #166, opened 2021-11-15
Add full path REST call

chemelli74:
Allow the caller to specify a full REST path (needed for Shelly Valve and Shelly Motion )

### DIFF
--- a/aioshelly/block_device.py
+++ b/aioshelly/block_device.py
@@ -218,7 +218,7 @@ class BlockDevice:
         if self.options.auth is None and self.requires_auth:
             raise AuthRequired
 
-        _LOGGER.debug("aiohttp request: /%s", path)
+        _LOGGER.debug("aiohttp request: /%s (params=%s)", path, params)
         resp: ClientResponse = await self.aiohttp_session.request(
             method,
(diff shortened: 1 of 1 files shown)


aioshelly/block_device.py
```diff
@@ -407,6 +407,10 @@ async def set_state(self, **kwargs: Any) -> dict[str, Any]:
             "get", f"{self.type}/{self.channel}", kwargs
         )
 
+    async def set_state_full_path(self, path: str, **kwargs: Any) -> dict[str, Any]:
```
thecode, line 410: This is now not really a function that sets a block state but a function that calls a generic REST command for the device.

I had the same comment from @balloob about RPC device that the name should match what it does.

In my opinion we should consider:
1. Name it something more related to what it does: `http_get_request` or `call_http_get`
2. Since it is not tied to a block now (unlike `set_state` which sets state within the path of a block) we should probably move it to the device class (same where `http_request` is located)

chemelli74: Actually I don't completely agree as this is just a different way to set the block.

chemelli74:
Changed approach, will rely on `device.http_request()` without the need for a change in this library.

Simone

PR #176, opened 2021-12-23
Add ident to RPC GetDeviceInfo call

chemelli74:
In order to get some more info about the device, the ident `param` is needed.

Reply will then have 2 more keys:

- batch
- key

I think it would be worth to decode key with one of https://jwt.io/libraries?language=Python
Decode content with current firmware looks like:

```json
{
  "iat": 1631078138,
  "mac": "xxxxxxxx",
  "m": "SPSW-004PE16EU",
  "b": "2132-002",
  "fp": "xxxxxx"
}

```
Simone

### DIFF
--- a/aioshelly/rpc_device.py
+++ b/aioshelly/rpc_device.py
@@ -126,7 +126,9 @@ class RpcDevice:
 
     async def update_device_info(self) -> None:
         """Get device info from 'Shelly.GetDeviceInfo'."""
-        self._device_info = await self._wsrpc.call("Shelly.GetDeviceInfo")
+        self._device_info = await self._wsrpc.call(
+            "Shelly.GetDeviceInfo", {"ident": True}
+        )
 
     async def update_config(self) -> None:
         """Get device config from 'Shelly.GetConfig'."""


thecode:
See comment here:
https://github.com/home-assistant/core/pull/62643#issuecomment-1000079760
I don't think we should add this implementation, it was left out by purpose, adding another PyPi requirement just for this field does not worth it. I would also suggest to verify again with the Shelly if this should be publicly available as it is not mentioned in the documentation and a dedicated flag was added to request this information.

chemelli74:
> See comment here: [home-assistant/core#62643 (comment)](https://github.com/home-assistant/core/pull/62643#issuecomment-1000079760) I don't think we should add this implementation, it was left out by purpose, adding another PyPi requirement just for this field does not worth it.

Fine for me.

> I would also suggest to verify again with the Shelly if this should be publicly available as it is not mentioned in the documentation and a dedicated flag was added to request this information.

Allterco devs are going to make those info publicly available, is just a matter of having time to write it down ;-)

Simone

chemelli74:
Main part of the Team decided to decline this functionality.

Simone

PR #180, opened 2021-12-30
Add methods to trigger self test/mute/unmute actions for Shelly Gas

bieniu:


aioshelly/block_device.py
```diff
@@ -256,6 +256,10 @@ async def trigger_reboot(self) -> None:
         """Trigger a device reboot."""
         await self.http_request("get", "reboot")
 
+    async def trigger_self_test(self) -> None:
+        """Trigger a Shelly Gas self test."""
```
thecode, line 260: Do we know of another device that has self test?
We should either name the function `trigger_shelly_gas_self_test` or rename the description to "Trigger device self test."

bieniu: Good point. Only Shelly Gas supports the self test so I'll change the method name.

thecode (approved):

chemelli74 (approved):

PR #185, opened 2022-01-24
Fix - prevent setting resolves during disconnect or at startup

thecode:
Prevent setting futures resolves during disconnect to prevent the following:
```
2022-01-05 20:05:21 ERROR (MainThread) [homeassistant] Error doing job: Task exception was never retrieved
  File "/usr/local/lib/python3.9/site-packages/aioshelly/wsrpc.py", line 165, in _rx_msgs
    self._handle_frame(frame)
  File "/usr/local/lib/python3.9/site-packages/aioshelly/wsrpc.py", line 151, in _handle_frame
    call.resolve.set_result(frame)
asyncio.exceptions.InvalidStateError: invalid state
```
May also happen during startup if we get a response from the device from previous connection.

### DIFF
--- a/aioshelly/wsrpc.py
+++ b/aioshelly/wsrpc.py
@@ -148,7 +148,8 @@ class WsRPC:
                 return
 
             call = self._calls.pop(frame_id)
-            call.resolve.set_result(frame)
+            if not call.resolve.cancelled():
+                call.resolve.set_result(frame)
 
         else:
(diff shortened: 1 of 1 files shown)


chemelli74 (approved):
LGTM!

thecode:
Offline discussion, it may also happen during startup if we get a response from the device from previous connection.

PR #189, opened 2022-02-05
Add more info to AttributeError

chemelli74:


thecode (approved):
Nice with adding the firmware

PR #191, opened 2022-02-08
Shelly Motion: enable/disable motion detection support.

mstefany:
Hi. I am missing some nice features for Shelly devices in Home Assistant, e.g. enabling/disabling Shelly Motion motion detection, setting motion/tamper sensitivity, or changing Shelly Gas Valve addon state. I would like to contribute those to aioshelly and Home Assistant Core. I created this small PR to get a feedback and see the whole process. Thank you.

### DIFF
--- a/aioshelly/block_device.py
+++ b/aioshelly/block_device.py
@@ -268,6 +268,12 @@ class BlockDevice:
         """Trigger a Shelly Gas unmute action."""
         await self.http_request("get", "unmute")
 
+    async def set_shelly_motion_detection(self, enable: bool) -> None:
+        """Enable or disable Shelly Motion motion detection."""
+        params = {"motion_enable": "true"} if enable else {"motion_enable": "false"}
+
+        await self.http_request("get", "settings", params)
+
     @property
     def requires_auth(self) -> bool:
         """Device check for authentication."""


aioshelly/block_device.py
```diff
@@ -268,6 +268,17 @@ async def trigger_shelly_gas_unmute(self) -> None:
         """Trigger a Shelly Gas unmute action."""
         await self.http_request("get", "unmute")
 
+    async def enable_shelly_motion_detection(self, enable: bool) -> None:
```
thecode, line 271:
```suggestion
    async def set_shelly_motion_detection(self, enable: bool) -> None:
```
To be consistent with other set commands we have

aioshelly/block_device.py
```diff
@@ -268,6 +268,17 @@ async def trigger_shelly_gas_unmute(self) -> None:
...
+        """Enable or disable Shelly Motion motion detection."""
+        params = {}
+
+        if enable:
+            params = {"motion_enable": "true"}
+        else:
+            params = {"motion_enable": "false"}
+
```
thecode, line 279:
```suggestion
        params = {"motion_enable": "true"} if enable else {"motion_enable": "false"}
```

thecode (changes requested):

aioshelly/block_device.py
mstefany: OK, changed, thank you.

aioshelly/block_device.py
mstefany: I was not sure if you write it in long or short form. Thank you for suggestion, I changed it.

thecode (approved):

bieniu (approved):
LGTM 👍

PR #192, opened 2022-02-14
shelly gen2 auth support

figadore:
This PR provides some of the logic needed to support Shelly gen2 digest authentication for websocket connections. See original issue at https://github.com/home-assistant/core/issues/63621

References:

- [Shelly docs on digest auth](https://shelly-api-docs.shelly.cloud/gen2/Overview/CommonDeviceTraits/#authentication-over-websocket)
- [typescript implementation](https://github.com/ALLTERCO/gen2-sample-code/blob/main/http-digest-auth/src/shellyHttpCall.ts)

I'm not super familiar with the structure of the rest of the codebase, so there's probably a lot here that needs to be restructured.

Anyway, the following code is what I used to verify that authentication is working for a local device

```python
import asyncio
from pprint import pprint

import aiohttp
import async_timeout

from aioshelly.common import ConnectionOptions
from aioshelly.rpc_device import RpcDevice


async def test_rpc_device():
    """Test Gen2 RPC (WebSocket) based device."""
    shelly_ip = "192.168.33.1"
    api_url = f"http://{shelly_ip}/rpc/Switch.Toggle?id=0"
    options = ConnectionOptions(shelly_ip, password="super-secret")

    async with aiohttp.ClientSession() as aiohttp_session:
        try:
            async with async_timeout.timeout(10):
                #device = await WsRPC.connect(aiohttp_session, options)
                device = await RpcDevice.create(aiohttp_session, options)
        except asyncio.TimeoutError:
            print("Timeout connecting to", ConnectionOptions.ip_address)
            return

        method = "Switch.Toggle"
        params = {"id": 0}
        response = await device.call_rpc(method, params)
        pprint("toggle response:")
        pprint(response)


if __name__ == "__main__":
    asyncio.run(test_rpc_device())
```

### DIFF
--- a/aioshelly/rpc_device.py
+++ b/aioshelly/rpc_device.py
@@ -22,7 +22,6 @@ def mergedicts(dict1: dict, dict2: dict) -> dict:
     return result
 
 
-
 class RpcDevice:
     """Shelly RPC device reppresentation."""
 
@@ -89,20 +88,12 @@ class RpcDevice:
         try:
             self.shelly = await get_info(self.aiohttp_session, self.options.ip_address)
 
-            if self.options.auth or not self.requires_auth:
-                await self._wsrpc.connect(self.aiohttp_session, self.options)
-                await asyncio.gather(
-                    self.update_device_info(),
-                    self.update_config(),
-                    self.update_status(),
-                )
-            elif self.requires_auth:
-                await self._wsrpc.connect(self.aiohttp_session, self.options.password)
-                await asyncio.gather(
-                    self.update_device_info(),
-                    self.update_config(),
-                    self.update_status(),
-                )
+            await self._wsrpc.connect(self.aiohttp_session, self.options)
+            await asyncio.gather(
+                self.update_device_info(),
+                self.update_config(),
+                self.update_status(),
+            )
             self.initialized = True
         finally:
             self._initializing = False
--- a/aioshelly/rpc_device.py
+++ b/aioshelly/rpc_device.py
@@ -4,15 +4,11 @@ from __future__ import annotations
 import asyncio
 from typing import Any, Callable, cast
 
-import hashlib
-import json
-import secrets
-
 import aiohttp
 from aiohttp.client import ClientSession
 
 from .common import ConnectionOptions, IpOrOptionsType, get_info, process_ip_or_options
-from .exceptions import AuthRequired, NotInitialized, WrongShellyGen, JSONRPCError
+from .exceptions import AuthRequired, NotInitialized, WrongShellyGen
 from .wsrpc import WsRPC
 
 
@@ -25,9 +21,6 @@ def mergedicts(dict1: dict, dict2: dict) -> dict:
             result[key] = mergedicts(dict1[key], value)
     return result
 
-
-def sha256(message):
-    return hashlib.sha256(message.encode('utf-8')).hexdigest()
 
 
 class RpcDevice:
@@ -97,46 +90,19 @@ class RpcDevice:
             self.shelly = await get_info(self.aiohttp_session, self.options.ip_address)
 
             if self.options.auth or not self.requires_auth:
-                await self._wsrpc.connect(self.aiohttp_session)
+                await self._wsrpc.connect(self.aiohttp_session, self.options)
                 await asyncio.gather(
                     self.update_device_info(),
                     self.update_config(),
                     self.update_status(),
                 )
             elif self.requires_auth:
-                try:
-                    await self._wsrpc.connect(self.aiohttp_session)
-                    result = await self._wsrpc.call("Sys.GetStatus", {})
-                except JSONRPCError as err:
-                    if err.code == 401:
-                        response_message = json.loads(err.message)
-
-                        username = "admin" #always
-                        password = "super-secret"
-                        ha1 = sha256(f"{username}:{response_message['realm']}:{password}")
-                        # Static noise
-                        ha2 = sha256("dummy_method:dummy_uri")
-                        cnonce = secrets.randbelow(10**8)
-                        if "nc" not in response_message:
-                            response_message['nc'] = 1
-                        hashed = sha256(f"{ha1}:{response_message['nonce']}:{response_message['nc']}:{cnonce}:auth:{ha2}")
-                        auth = {
-                                "realm": response_message["realm"],
-                                "username": "admin",
-                                "nonce": response_message["nonce"],
-                                "cnonce": cnonce,
-                                "response": hashed,
-                                "algorithm": "SHA-256"
-                        }
-                        self._wsrpc.set_auth(auth)
-                        # Verify auth
-                        try:
-                            new_result = await self._wsrpc.call("Sys.GetStatus")
-                        except JSONRPCError as err:
-                            print("Auth still failed, password may be incorrect")
-                            raise err
-                    else:
-                        raise err
+                await self._wsrpc.connect(self.aiohttp_session, self.options.password)
+                await asyncio.gather(
+                    self.update_device_info(),
+                    self.update_config(),
+                    self.update_status(),
+                )
             self.initialized = True
         finally:
             self._initializing = False
--- a/aioshelly/rpc_device.py
+++ b/aioshelly/rpc_device.py
@@ -4,11 +4,15 @@ from __future__ import annotations
 import asyncio
 from typing import Any, Callable, cast
 
+import hashlib
+import json
+import secrets
+
 import aiohttp
 from aiohttp.client import ClientSession
 
 from .common import ConnectionOptions, IpOrOptionsType, get_info, process_ip_or_options
-from .exceptions import AuthRequired, NotInitialized, WrongShellyGen
+from .exceptions import AuthRequired, NotInitialized, WrongShellyGen, JSONRPCError
 from .wsrpc import WsRPC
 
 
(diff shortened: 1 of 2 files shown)


thecode:
All the authorization logic should be handled in [wsrpc](https://github.com/home-assistant-libs/aioshelly/blob/main/aioshelly/wsrpc.py) and not in `rpc_device`. 

The authorization parameters are pass using the `ConnectionOptions` class:
https://github.com/home-assistant-libs/aioshelly/blob/25296175e96df9e634e9ff9ed3a6f4a68e9410dd/aioshelly/common.py#L20-L28

Please look at the example how we use it for Gen1 devices: https://github.com/home-assistant-libs/aioshelly/blob/main/example.py

figadore:
Took another stab at it

> All the authorization logic should be handled in [wsrpc](https://github.com/home-assistant-libs/aioshelly/blob/main/aioshelly/wsrpc.py) and not in `rpc_device`.

Moved

> The authorization parameters are pass using the `ConnectionOptions` class:

Passing password in using ConnectionOptions

chemelli74:
If I'm not wrong, there should be a way to know if device requires auth without doing a try/except.

Simone

thecode:
> If I'm not wrong, there should be a way to know if device requires auth without doing a try/except.
> 
> Simone

We do, same as for Gen1, @figadore I highly suggest to look at how we implement it for Gen1. `/shelly` should include a flag if auth required.

figadore:
> If I'm not wrong, there should be a way to know if device requires auth without doing a try/except.

https://shelly-api-docs.shelly.cloud/gen2/Overview/CommonDeviceTraits/#authentication-process

I tried to get this to work with just the `requires_auth` property, but I can't find a way. The [docs](https://shelly-api-docs.shelly.cloud/gen2/Overview/CommonDeviceTraits/#authentication-process) seem to indicate that you get the challenge parameters from the first 401 response. If you know how else this can be accomplished, let me know

aioshelly/wsrpc.py
```diff
@@ -38,15 +42,23 @@ class RPCCall:
...
         """Initialize RPC class."""
         self.call_id = call_id
-        self.params = params
         self.method = method
+        self.params = params
         self.src = route.src
         self.dst = route.dst
         self.resolve: asyncio.Future = asyncio.Future()
+        self.auth = {}
+        if params and "auth" in params:
```
figadore, line 59: not sure this is the best way to accomplish handling auth... mostly trying to avoid pylint's "too-many-arguments"

figadore:
@thecode , @chemelli74, I wasn't able to find the auth challenge parameters in the initial request. The `requires_auth` property of RpcDevice can be incorporated somewhere if needed, but it still looks like this won't help to avoid doing a try/catch to get the params needed to construct the `auth` property for future requests. Let me know if I'm missing anything, or if there's anything else I can do to improve this PR

aioshelly/wsrpc.py
thecode: We should not do workarounds to bypass pylint, we can either increase this limit in pylint or add it to the route class.
If it is added to the route class the class will be needed to be renamed to a better name, such as `ConnectionData`

thecode:
> @thecode , @chemelli74, I wasn't able to find the auth challenge parameters in the initial request. The `requires_auth` property of RpcDevice can be incorporated somewhere if needed, but it still looks like this won't help to avoid doing a try/catch to get the params needed to construct the `auth` property for future requests. Let me know if I'm missing anything, or if there's anything else I can do to improve this PR

Check the the data in `/shelly` or look at the example I linked for Gen1

figadore:
> Check the the data in `/shelly` 

@thecode What/where is `/shelly`?

> look at the example I linked for Gen1

Gen 1's BlockDevice class uses `self.options.auth` and `self.requires_auth` (e.g. [here](https://github.com/home-assistant-libs/aioshelly/blob/7c879190a44499155a4e992ccc7b1d3790f30660/aioshelly/block_device.py#L97)), but I don't think either of these can be used to prevent an initial request that produces a 401, if that's what you're suggesting. The workflow as described by the Shelly docs is as follows. Note that there is no way (that I know of) to provide credentials in 3. without getting the response from 2. at least once.

```
Steps in the process:

1. Client requests a protected resource without providing credentials.
2. Server response containing error 401 (unauthorized) is received.
3. Client requests the same protected resource but this time providing credentials.
4. The request is successful and access to the resource is granted.
```

If it's not what you're suggesting, I'm all for consistency, so what other areas should be changed so the Gen2 auth lines up more closely with Gen 1?

aioshelly/wsrpc.py
```diff
@@ -79,7 +101,11 @@ def _next_id(self) -> int:
         self._call_id += 1
         return self._call_id
 
-    async def connect(self, aiohttp_session: aiohttp.ClientSession) -> None:
+    async def connect(
+        self,
+        aiohttp_session: aiohttp.ClientSession,
+        options: ConnectionOptions | None = None,
```
kzyapkov, line 107: `options` here contains the `ip_address` already passed in the constructor. It isn't obvious that it is being ignored in the `connect()` method, along with the other properties of the `ConnectionOptions` instance.

If it were me, I'd simply pass an optional `password` in the constructor, and complicate it further if needed in the future.

aioshelly/wsrpc.py
```diff
@@ -99,6 +125,44 @@ async def connect(self, aiohttp_session: aiohttp.ClientSession) -> None:
 
         _LOGGER.info("Connected to %s", self._ip_address)
 
+        # https://shelly-api-docs.shelly.cloud/gen2/Overview/CommonDeviceTraits/#authentication-over-websocket
+        if options and options.password:
+            try:
+                # First API call should fail with 401 if auth enabled
+                await self.call("Sys.GetStatus", {})
```
kzyapkov, line 132: I think the `WsRPC` class should abstain from invoking specific methods on the device -- it should focus on purely providing the RPC channel. The authentication flow can happen for the first (or every?) call which gets a 401. That way, `WsRPC` doesn't need to know about methods and their semantics, and will do the authentication step only when necessary.

aioshelly/wsrpc.py
```diff
@@ -204,13 +268,24 @@ def connected(self) -> bool:
...
         self, method: str, params: dict[str, Any] | None = None, timeout: int = 10
     ) -> dict[str, Any]:
         """Websocket RPC call."""
         if self._client is None:
             raise RuntimeError("Not connected")
 
+        if self._auth:
+            if not params:
+                params = {}
+            if "auth" not in params:
```
kzyapkov, line 287: `auth` should not be in params. It should be treated separately. For one, it is not impossible for a method to require an `auth` parameter.

aioshelly/rpc_device.py
```diff
@@ -88,13 +88,12 @@ async def initialize(self) -> None:
...
 
-            if self.options.auth or not self.requires_auth:
-                await self._wsrpc.connect(self.aiohttp_session)
-                await asyncio.gather(
-                    self.update_device_info(),
-                    self.update_config(),
-                    self.update_status(),
-                )
+            await self._wsrpc.connect(self.aiohttp_session, self.options)
+            await asyncio.gather(
```
kzyapkov, line 92: Instead of doing the auth dance in `WsRPC.connect()` I would reorganize this flow instead:
 * fetch device info
 * check if `auth_en`, if yes:
   * password is unknown -- error out.
   * password is known -- make sure WsRPC knows about it, possibly make it send the `auth` data without getting a 401 error first
   * proffit

kzyapkov (commented):
@figadroe thanks for working on this!

I admit, our documentation on the topic isn't the best, which we should fix sooner than later. Other than that, [here](https://gist.github.com/kzyapkov/4e425ba99b2347f56307dbaffc1363fb) is a small demo to show how to auth without getting the challenge first, with pointers to source and docs upstream. To make it work, I commented out the "auth magic" part from `WsRPC.connect()`. `nonce` will expire every once in a while, so it's important to have the auth mechanism available for whenever a 401 with challenge gets back.

aioshelly/wsrpc.py
thecode: That is exactly what I was looking for, similar what we do for Gen1, if we have auth we should just pass it on every call not trying to fail first and retry, this with the example how to make the first auth https://github.com/home-assistant-libs/aioshelly/pull/192#pullrequestreview-899520621 would provide a clean solution.
@figadore let me know if you plan to work on this changes, otherwise I can try to pick it up and we commit it together.

chemelli74:
> > Check the the data in `/shelly`
> 
> @thecode What/where is `/shelly`?

Check `auth_en` from http://ip_address/shelly:

```json
{
  "name": "Test 1Plus",
  "id": "shellyplus1-**redacted**",
  "mac": "**redacted**",
  "model": "SNSW-001X16EU",
  "gen": 2,
  "fw_id": "20220308-100547/0.10.0-g4eda0ff",
  "ver": "0.10.0",
  "app": "Plus1",
  "auth_en": false,
  "auth_domain": null
}
```

figadore:
> @figadore thanks for working on this!
> 

Thanks for the feedback. I ended up revamping my home network to accommodate the lack of auth support in gen 2 Shelly, so my immediate need for this has been reduced. Additionally, my life has become busy again, but I do plan on getting back to this in the coming weeks. Meanwhile, I'm definitely not against anyone hijacking this PR to bring it to completion (or even scrapping it and starting over)

thecode:
Replaced by #209

thecode:
Replaced by #209

PR #196, opened 2022-02-28
Add missing model names

bieniu:


thecode (approved):

aioshelly/const.py
```diff
@@ -46,7 +47,10 @@
     # Gen2 RPC based models
     "SNSW-001P16EU": "Shelly Plus 1PM",
     "SNSW-001X16EU": "Shelly Plus 1",
+    "SNSW-0024X": "Shelly Plus i4",
```
kzyapkov, line 50: We officially call it I4, with a capital `I`, not that it matters much.

bieniu: @kzyapkov In this case, the name in the Shelly Shop is wrong.

Screenshot_20220228-204535

PR #209, opened 2022-04-04
Add Shelly gen2 authentication support

thecode:
Implement Shelly gen2 authentication support, continue work by @figadore https://github.com/home-assistant-libs/aioshelly/pull/192

Thanks @figadore and @kzyapkov

### DIFF
--- a/aioshelly/rpc_device.py
+++ b/aioshelly/rpc_device.py
@@ -2,6 +2,8 @@
 from __future__ import annotations
 
 import asyncio
+import hashlib
+import time
 from typing import Any, Callable, cast
 
 import aiohttp
@@ -10,6 +12,11 @@ from aiohttp.client import ClientSession
 from .common import ConnectionOptions, IpOrOptionsType, get_info, process_ip_or_options
 from .exceptions import AuthRequired, NotInitialized, WrongShellyGen
 from .wsrpc import WsRPC
+
+
+def hex_hash(message: str) -> str:
+    """Get hex representation of sha256 hash of string."""
+    return hashlib.sha256(message.encode("utf-8")).hexdigest()
 
 
 def mergedicts(dict1: dict, dict2: dict) -> dict:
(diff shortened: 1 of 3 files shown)


aioshelly/wsrpc.py
```diff
@@ -25,26 +29,68 @@
...
+    """RPC Auth data class."""
+
+    realm: str
+    username: str
+    password: str
+
+    def __post_init__(self) -> None:
+        """Call after initialization."""
+        self.ha1 = hex_hash(f"{self.username}:{self.realm}:{self.password}")
+        self.ha2 = hex_hash("dummy_method:dummy_uri")
```
balloob, line 48: why generate this on `__post_init__` and not just statically set it? 

Let's add a comment why this is necessary.

aioshelly/wsrpc.py
```diff
@@ -206,34 +258,62 @@ def connected(self) -> bool:
...
+
+        if code != 401:
+            raise JSONRPCError(code, msg)
+        if self._auth_data is None:
+            raise InvalidAuthError(code, msg)
+
+        # Update auth from response and try with new auth data
+        auth = json.loads(msg)
+        self._session.auth = self._auth_data.get_auth(auth["nonce"], auth.get("nc", 1))
+        resp = await self._rpc_call(method, params, timeout)
```
balloob, line 281: Better to call yourself again but now set a keyword arg to not refresh auth. Easier to keep the code in sync.

```suggestion
        return await self.call(method, params, timeout, handle_auth=False)
```

balloob (approved):

aioshelly/rpc_device.py
```diff
@@ -88,13 +88,22 @@ async def initialize(self) -> None:
...
-                    self.update_device_info(),
-                    self.update_config(),
-                    self.update_status(),
+            if self.requires_auth:
+                if not self.options.auth:
+                    raise AuthRequired
+
+                self._wsrpc.set_auth_data(
+                    self.shelly["auth_domain"],
+                    cast(str, self.options.username),
```
balloob, line 97: Why does this need casting?

balloob (approved):

aioshelly/rpc_device.py
thecode: I used the `self.options.auth`, this is why it confuses mypy, but there is no reason to use it here, changed it to 
https://github.com/home-assistant-libs/aioshelly/blob/00b286b4997ac176633d64b44ac2a51cf7273d3b/aioshelly/rpc_device.py#L92

balloob (approved):

PR #244, opened 2022-08-02
Add inbound websocket server for Gen2 sleeping devices

thecode:
## Add websocket server to enable Shelly Gen2 sleeping devices to send back data.

This PR copy most of the logic we use in CoAP for Gen1:
- Single server with shared context
- Init device upon receiving a message from the device
- Subscribe for push updates by device IP

To test this on a sleeping device:
- In the device web under `Networks` -> `Outbound websocket` enable the outbound websocket and set URL that points to the host: `ws://<host_ip>:5683/` (note that the `/` at the end is mandatory)
- Run the example : `example.py --ip <device_ip> -g2`
Wait for an update or push the button on the device (when pushing back to sleep device also sends an update)

## Breaking change
`RpcDevice.create` requires an additional parameter for the shared inbound websocket server

When creating a device the following:
```python
RpcDevice.create(aiohttp_session, options, init)
```
Should be changed to:
```python
ws_context = WsServer()
await ws_context.initialize(ws_port)
RpcDevice.create(aiohttp_session, ws_context, options, init)
```

### DIFF
--- a/aioshelly/block_device.py
+++ b/aioshelly/block_device.py
@@ -7,11 +7,11 @@ from typing import Any, Callable, cast
 
 import aiohttp
 import async_timeout
-from aiohttp.client_reqrep import ClientResponse
+from aiohttp.client import ClientResponse
 
 from .coap import COAP, CoapMessage
 from .common import ConnectionOptions, IpOrOptionsType, get_info, process_ip_or_options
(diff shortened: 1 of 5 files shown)


bieniu (approved):
I tested this change with Plus HT and it's working well. LGTM 👍

chemelli74:
Quick test result is ok. Good job!

chemelli74 (approved):

PR #254, opened 2022-09-21
Suppress `ClientResponseError` in `update_status()` and `update_settings()`

bieniu:
To avoid `ClientResponseError` exception when the user changes the password for the sleeping device, suppress the exception is necessary.

Related to: https://github.com/home-assistant/core/pull/78786

### DIFF
--- a/aioshelly/block_device.py
+++ b/aioshelly/block_device.py
@@ -2,11 +2,13 @@
 from __future__ import annotations
 
 import asyncio
+import contextlib
 import logging
 from typing import Any, Callable, cast
 
 import aiohttp
 import async_timeout
+from aiohttp import ClientResponseError
 from aiohttp.client import ClientResponse
 
 from .coap import COAP, CoapMessage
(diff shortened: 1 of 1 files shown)


thecode (approved):

aioshelly/block_device.py
```diff
@@ -195,11 +197,13 @@ def subscribe_updates(self, update_listener: Callable) -> None:
 
     async def update_status(self) -> None:
         """Device update from /status (HTTP)."""
-        self._status = await self.http_request("get", "status")
+        with contextlib.suppress(ClientResponseError):
```
MartinHjelmare, line 200: Is it not interesting for the library user to know that there was an error?

bieniu: `ClientResponseError` exception is raised here when initializing the battery powered device when the user has changed the password. Init happens on a task, so we are unable to catch this exception. In other cases, `settings` and `status` properties are protected and raise `NotInitialized` or `AuthRequired` exceptions.
https://github.com/home-assistant-libs/aioshelly/blob/b4ad8b776ffa99b8580645edafd460139f1dd3d2/aioshelly/block_device.py#L291-L311

thecode: I agree with @MartinHjelmare we should at least log it as debug (we can't raise here since this code is called on a background task). We can also check the error code in the `ClientResponseError`.
However since this is a patch on v2 releases I would not change it now. We are moving to v3 soon (there is already a 3.0.0 release), we can improve this when adding this into the 3.x release.

bieniu: Sure, we can log this exception.

PR #268, opened 2022-10-21
Add RpcCallError exception

thecode:
Noted by @kzyapkov, if the RPC call contains wrong/incomplete parameters or device could not execute the call for some reason but managed to reply back we have no reason to destroy the connection with the device.

This PR adds a new exception `RpcCallError` which will be raised upon failures to `call_rpc`, However during device initialization if the there is a failure, `DeviceConnectionError` is still raised since in this case this is not due to user or parameter error and we can't continue the device initialization.

### DIFF
--- a/aioshelly/const.py
+++ b/aioshelly/const.py
@@ -3,16 +3,13 @@ import asyncio
 
 import aiohttp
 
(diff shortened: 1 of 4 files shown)


kzyapkov:
LGTM and works well for my use case.

bieniu (approved):
LGTM 👍

PR #269, opened 2022-10-22
Add set thermostat state (Shelly TRV)

thecode:
Add method to set thermostat state (Shelly TRV), to avoid using direct HTTP request in Home Assistant


bieniu (approved):
LGTM 👍

thecode:
Thanks!

PR #282, opened 2022-11-14
Improve JSON performance

bdraco:
with #281 we now process many more websocket
messages. As we do not use the Home Assistant
aiohttp session we need to use orjson manually
to avoid the significant json overhead

### DIFF
--- a/aioshelly/json.py
+++ b/aioshelly/json.py
@@ -4,7 +4,7 @@ from typing import Any
 
 import orjson
 
-json_loads = orjson.loads
+json_loads = orjson.loads  # pylint: disable=no-member
 
 
 def json_encoder_default(obj: Any) -> Any:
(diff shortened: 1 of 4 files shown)


aioshelly/json.py
```diff
@@ -0,0 +1,32 @@
...
+    orjson supports serializing dataclasses natively which
+    eliminates the need to implement as_dict in many places
+    when the data is already in a dataclass. This works
+    well as long as all the data in the dataclass can also
+    be serialized.
+
+    If it turns out to be a problem we can disable this
+    with option |= orjson.OPT_PASSTHROUGH_DATACLASS and it
+    will fallback to as_dict
+    """
```
bdraco, line 29:
```suggestion
    """Dump json string."""
```

balloob (approved):

PR #288, opened 2022-11-17
Device refactor

thecode:
Move devices into their own folders so we can also split the models and APIs to different files.
For now I only moved the files with minimal code changes, moving code to other files will be done in follow-up PRs.
Added `find_packages` to include all sub-packages and verified that the build wheel contain all and only the needed files.


bdraco (approved):
LGTM

bieniu (approved):
LGTM 👍

PR #294, opened 2022-12-08
Fix pre-commit flake8

thecode:
`pre-commit` used old `flake8` version from `gitlab` which is no longer available, updated to the latest one using the recommended repo from the docs:
https://flake8.pycqa.org/en/latest/user/using-hooks.html#usage-with-the-pre-commit-git-hooks-framework


frenck (approved):

thecode:
Thanks!

PR #307, opened 2022-12-12
Fix state updates never coming back after reconnect

bdraco:
If we hit a connection or invalid auth error we would clear the _update_listener because it called shutdown which cleared the listener and closed the websocket connection.

We now only shutdown the websocket connection so
updates keep working once we reconnect


bdraco:
Confirmed this fixes it.

We should do another patch release with this fix.

bdraco:
Testing:

unplug shelly device.

Wait 2-3 minutes to make sure we see the disconnection.

Plug back in device.

Should recover within 10s now

balloob (approved):

PR #310, opened 2022-12-14
5.2.0

bdraco:


bdraco:
I had to do some manually edits to make everything work since release drafter got a bit confused with the 2 cherry-pick publishes but should be all sorted out now https://github.com/home-assistant-libs/aioshelly/releases/tag/5.2.0

PR #314, opened 2022-12-21
Reduce pressure on shelly device by increasing receive buffer

bdraco:
If there are a lot of bluetooth packets coming over the wire we want the shelly to be able to empty its buffer as fast as possible if the event loop is busy so we increase the buffer size a bit to reduce the memory pressure on the shelly


thecode (approved):

bdraco:
Thanks

PR #322, opened 2023-01-05
Accept IP address as-is if it appears to be IPv4 with port specified

GarrStau:
Allow an IPv4 with port like "192.168.0.203:8002" to be used as IP address.
In my instance, this allows connecting and controlling a device connected to another Shelly through the Range Extender feature.

### DIFF
--- a/.github/workflows/release-drafter.yml
+++ b/.github/workflows/release-drafter.yml
@@ -10,6 +10,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       # Drafts your next Release notes as Pull Requests are merged into "master"
-      - uses: release-drafter/release-drafter@v5.21.1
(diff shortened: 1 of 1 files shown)


thecode (changes requested):
I prefer not use `regex` it would be better to add a `port` parameter with a default value.
However, note that I tested such change in the past and it is not enough to allow different ports, there are multiple scenarios which needed to be handled correctly for CoAP periodic updates (Gen1) or outbound websocket (Gen2)

GarrStau:
@thecode I am attempting to get a developer environment set up, and will make a new PR when I have a fix with "port" added to the aioshelly.common.ConnectionOptions class.

I have access to a Shelly Plug US and a Shelly Motion 2, which I intend to use to troubleshoot and verify.

PR #325, opened 2023-01-07
Fix broken callback in example.py

GarrStau:
example.py was broken for Gen2 devices in https://github.com/home-assistant-libs/aioshelly/pull/283/files

Callback function needs to accept 2 parameters for Gen 2 devices but 1 parameter for Gen 1 devices - there are a lot of ways to fix this, if you'd prefer to see something else (print status? keyword arg instead of *args?) I can change it.

### DIFF
--- a/example.py
+++ b/example.py
@@ -26,7 +26,7 @@ from aioshelly.exceptions import (
     ShellyError,
     WrongShellyGen,
 )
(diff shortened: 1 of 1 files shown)


GarrStau:
... 2 minutes too slow

PR #326, opened 2023-01-08
Add support for specifying port in Gen2 devices

GarrStau:
(Will rebase when https://github.com/home-assistant-libs/aioshelly/pull/324 is merged)

Enables support for Range Extender connected Gen2 devices (following request to use port as a parameter in https://github.com/home-assistant-libs/aioshelly/pull/322).

Unfortunately, it looks like CoAP is not possible due to Range Extender only forwarding to port 80 on the device. If my understanding is incorrect I'd be happy to try and incorporate support for Gen1 devices as well.

If this PR is accepted, there will need to be additional work to the Home Assistant configuration flow to specify ports.

### DIFF
--- a/aioshelly/ble/__init__.py
+++ b/aioshelly/ble/__init__.py
@@ -92,7 +92,7 @@ def create_scanner(
         name,
         HaBluetoothConnector(
             # no active connections to shelly yet
-            client=None,  # type: ignore[arg-type]
+            client=None,
             source=source,
             can_connect=lambda: False,
         ),
@@ -114,6 +114,6 @@ async def async_ensure_ble_enabled(device: RpcDevice) -> bool:
     ble_enable = await device.ble_setconfig(enable=True, enable_rpc=True)
     if not ble_enable["restart_required"]:
         return False
-    LOGGER.info("BLE enabled, restarting device %s", device.ip_address)
+    LOGGER.info("BLE enabled, restarting device %s:%s", device.ip_address, device.port)
     await device.trigger_reboot(3500)
     return True
--- a/aioshelly/ble/__init__.py
+++ b/aioshelly/ble/__init__.py
@@ -92,7 +92,7 @@ def create_scanner(
         name,
         HaBluetoothConnector(
             # no active connections to shelly yet
-            client=None,
+            client=None,  # type: ignore[arg-type]
             source=source,
             can_connect=lambda: False,
         ),
--- a/aioshelly/block_device/device.py
+++ b/aioshelly/block_device/device.py
@@ -18,6 +18,7 @@ from yarl import URL
 from ..common import ConnectionOptions, IpOrOptionsType, get_info, process_ip_or_options
 from ..const import CONNECT_ERRORS, DEVICE_IO_TIMEOUT, HTTP_CALL_TIMEOUT, MODEL_RGBW2
 from ..exceptions import (
+    CustomPortNotSupported,
     DeviceConnectionError,
     FirmwareUnsupported,
     InvalidAuthError,
@@ -117,6 +118,10 @@ class BlockDevice:
         """Device initialization."""
         if self._initializing:
             raise RuntimeError("Already initializing")
+
+        # GEN1 cannot be configured behind a range extender as CoAP port cannot be natted
+        if self.options.port != 80:
+            raise CustomPortNotSupported
 
         self._initializing = True
         self.initialized = False
--- a/aioshelly/common.py
+++ b/aioshelly/common.py
@@ -44,6 +44,7 @@ class ConnectionOptions:
     temperature_unit: str = "C"
     auth: aiohttp.BasicAuth | None = None
     device_mac: str | None = None
+    port: int = 80
 
     def __post_init__(self) -> None:
         """Call after initialization."""
@@ -81,29 +82,30 @@ async def get_info(
     aiohttp_session: aiohttp.ClientSession,
     ip_address: str,
     device_mac: str | None = None,
+    port: int = 80,
 ) -> dict[str, Any]:
     """Get info from device through REST call."""
     try:
         async with aiohttp_session.get(
-            URL.build(scheme="http", host=ip_address, path="/shelly"),
+            URL.build(scheme="http", host=ip_address, port=port, path="/shelly"),
             raise_for_status=True,
             timeout=DEVICE_IO_TIMEOUT,
         ) as resp:
             result: dict[str, Any] = await resp.json()
     except CONNECT_ERRORS as err:
         error = DeviceConnectionError(err)
-        _LOGGER.debug("host %s: error: %r", ip_address, error)
+        _LOGGER.debug("host %s:%s: error: %r", ip_address, port, error)
         raise error from err
 
     mac = result["mac"]
     if device_mac and device_mac != mac:
         mac_err = MacAddressMismatchError(f"Input MAC: {device_mac}, Shelly MAC: {mac}")
-        _LOGGER.debug("host %s: error: %r", ip_address, mac_err)
+        _LOGGER.debug("host %s:%s: error: %r", ip_address, port, mac_err)
         raise mac_err
 
     if not shelly_supported_firmware(result):
         fw_error = FirmwareUnsupported(result)
-        _LOGGER.debug("host %s: error: %r", ip_address, fw_error)
+        _LOGGER.debug("host %s:%s: error: %r", ip_address, port, fw_error)
         raise fw_error
 
     return result
--- a/aioshelly/rpc_device/device.py
+++ b/aioshelly/rpc_device/device.py
@@ -73,7 +73,9 @@ class RpcDevice:
         self._status: dict[str, Any] | None = None
         self._event: dict[str, Any] | None = None
         self._config: dict[str, Any] | None = None
-        self._wsrpc = WsRPC(options.ip_address, self._on_notification)
+        self._wsrpc = WsRPC(
+            options.ip_address, self._on_notification, port=options.port
+        )
         sub_id = options.ip_address
         if options.device_mac:
             sub_id = options.device_mac
@@ -138,6 +140,11 @@ class RpcDevice:
         """Device ip address."""
         return self.options.ip_address
 
+    @property
+    def port(self) -> int:
+        """Device port."""
+        return self.options.port
+
     async def initialize(self, async_init: bool = False) -> None:
         """Device initialization."""
         if self._initializing:
@@ -148,7 +155,10 @@ class RpcDevice:
         ip = self.options.ip_address
         try:
             self._shelly = await get_info(
-                self.aiohttp_session, self.options.ip_address, self.options.device_mac
+                self.aiohttp_session,
+                self.options.ip_address,
+                self.options.device_mac,
+                self.options.port,
             )
 
             if self.requires_auth:
@@ -171,7 +181,7 @@ class RpcDevice:
             self.initialized = True
         except InvalidAuthError as err:
             self._last_error = InvalidAuthError(err)
-            _LOGGER.debug("host %s: error: %r", ip, self._last_error)
+            _LOGGER.debug("host %s:%s: error: %r", ip, self.port, self._last_error)
             # Auth error during async init, used by sleeping devices
             # Will raise 'invalidAuthError' on next property read
             if not async_init:
@@ -180,13 +190,13 @@ class RpcDevice:
             self.initialized = True
         except (MacAddressMismatchError, FirmwareUnsupported) as err:
             self._last_error = err
-            _LOGGER.debug("host %s: error: %r", ip, err)
+            _LOGGER.debug("host %s:%s: error: %r", ip, self.port, err)
             if not async_init:
                 await self._disconnect_websocket()
                 raise
         except (*CONNECT_ERRORS, RpcCallError) as err:
             self._last_error = DeviceConnectionError(err)
-            _LOGGER.debug("host %s: error: %r", ip, self._last_error)
+            _LOGGER.debug("host %s:%s: error: %r", ip, self.port, self._last_error)
             if not async_init:
                 await self._disconnect_websocket()
                 raise DeviceConnectionError(err) from err
--- a/aioshelly/rpc_device/device.py
+++ b/aioshelly/rpc_device/device.py
@@ -153,6 +153,7 @@ class RpcDevice:
         self._initializing = True
         self.initialized = False
         ip = self.options.ip_address
+        port = self.options.port
         try:
             self._shelly = await get_info(
                 self.aiohttp_session,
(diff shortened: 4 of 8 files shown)


aioshelly/common.py
```diff
@@ -67,24 +68,25 @@ async def process_ip_or_options(ip_or_options: IpOrOptionsType) -> ConnectionOpt
 
 
 async def get_info(
-    aiohttp_session: aiohttp.ClientSession, ip_address: str
+    aiohttp_session: aiohttp.ClientSession, ip_address: str, port: str | None = None
```
thecode, line 71:
```suggestion
    aiohttp_session: aiohttp.ClientSession, ip_address: str, port: int| None = 80
```

aioshelly/common.py
```diff
@@ -67,24 +68,25 @@ async def process_ip_or_options(ip_or_options: IpOrOptionsType) -> ConnectionOpt
...
 ) -> dict[str, Any]:
     """Get info from device through REST call."""
+    opt_port = f":{port}" if port else ""
```
thecode, line 74:
```suggestion
```

aioshelly/common.py
```diff
@@ -67,24 +68,25 @@ async def process_ip_or_options(ip_or_options: IpOrOptionsType) -> ConnectionOpt
...
-            f"http://{ip_address}/shelly",
+            f"http://{ip_address}{opt_port}/shelly",
             raise_for_status=True,
             timeout=DEVICE_IO_TIMEOUT,
         ) as resp:
             result: dict[str, Any] = await resp.json()
     except CONNECT_ERRORS as err:
         error = DeviceConnectionError(err)
-        _LOGGER.debug("host %s: error: %r", ip_address, error)
+        _LOGGER.debug("host %s%s: error: %r", ip_address, opt_port, error)
```
thecode, line 84:
```suggestion
        _LOGGER.debug("host %s:%s: error: %r", ip_address, port, error)
```

aioshelly/common.py
```diff
@@ -67,24 +68,25 @@ async def process_ip_or_options(ip_or_options: IpOrOptionsType) -> ConnectionOpt
...
 async def get_info(
-    aiohttp_session: aiohttp.ClientSession, ip_address: str
+    aiohttp_session: aiohttp.ClientSession, ip_address: str, port: str | None = None
 ) -> dict[str, Any]:
     """Get info from device through REST call."""
+    opt_port = f":{port}" if port else ""
     try:
         async with aiohttp_session.get(
-            f"http://{ip_address}/shelly",
+            f"http://{ip_address}{opt_port}/shelly",
```
thecode, line 77:
```suggestion
            f"http://{ip_address}:{port}/shelly",
```

aioshelly/common.py
```diff
@@ -67,24 +68,25 @@ async def process_ip_or_options(ip_or_options: IpOrOptionsType) -> ConnectionOpt
...
     except CONNECT_ERRORS as err:
         error = DeviceConnectionError(err)
-        _LOGGER.debug("host %s: error: %r", ip_address, error)
+        _LOGGER.debug("host %s%s: error: %r", ip_address, opt_port, error)
         raise error from err
 
     if not shelly_supported_firmware(result):
         fw_error = FirmwareUnsupported(result)
-        _LOGGER.debug("host %s: error: %r", ip_address, fw_error)
+        _LOGGER.debug("host %s%s: error: %r", ip_address, opt_port, fw_error)
```
thecode, line 89:
```suggestion
        _LOGGER.debug("host %s:%s: error: %r", ip_address, port, fw_error)
```

thecode (dismissed):
Please make the `port` and integer with a default value.

GarrStau:
> Please make the `port` and integer with a default value.

I hope I was correct in assuming you wanted this done throughout the library.

GarrStau:
I also realized the example would break when `BlockDevice.port` was printed, so I went ahead and added it to the HTTP calls. I don't have a way to test this, however.

aioshelly/rpc_device/wsrpc.py
```diff
@@ -404,7 +418,7 @@ async def _rpc_call(
 
     async def _send_json(self, data: dict[str, Any]) -> None:
         """Send json frame to device."""
-        _LOGGER.debug("send(%s): %s", self._ip_address, data)
+        _LOGGER.debug("send(%s%s): %s", self._ip_address, self._opt_port, data)
```
bieniu, line 421:
```suggestion
        _LOGGER.debug("send(%s%s): %s", self._ip_address, self._port, data)
```

bieniu:
I tested this via example.py and in HA and everything seems to work. I managed to configure the device connected via the range extender. @thecode what do you think about this change? Any concerns?

chemelli74 (approved):
Works fine here. Thx !

thecode:
> I tested this via example.py and in HA and everything seems to work. I managed to configure the device connected via the range extender. @thecode what do you think about this change? Any concerns?

My main concern is that we need to support both Gen1 & Gen2, does it work on Gen1? (PR title is "Added support for specifying port in Gen2 devices")

bieniu:
> does it work on Gen1?

Good point, I didn't test gen1 devices.

chemelli74:
> > I tested this via example.py and in HA and everything seems to work. I managed to configure the device connected via the range extender. @thecode what do you think about this change? Any concerns?
> 
> 
> 
> My main concern is that we need to support both Gen1 & Gen2, does it work on Gen1? (PR title is "Added support for specifying port in Gen2 devices")

Am I wrong or gen1 doesn't support proxy ?

thecode:
> Am I wrong or gen1 doesn't support proxy ?

This PR doesn't add proxy support it adds support for proxied devices (e.g. Devices connected via another device), For Shelly cloud you can proxy Gen1 device using Gen2 in extender mode, if we add support I think we should add both so we don't get issues that some devices are not working. I already added support for incoming CoAP via device MAC so it should work in theory for Gen1 that are connected via extender but I don't have time to test it now.

bieniu:
I think that before merging we need to have some kind of PoC for integration and test if this change doesn't break something.

Two things need to be checked:
- whether zeroconf packets go from the device to the main lan
- what if the IP address of the device behind the range extender changes and zeroconf packets do not go through

GarrStau:
> I think that before merging we need to have some kind of PoC for integration and test if this change doesn't break something.
> 
> Two things need to be checked:
> 
> * whether zeroconf packets go from the device to the main lan
> * what if the IP address of the device behind the range extender changes and zeroconf packets do not go through

I am not well versed in any of these technologies, and I wouldn't trust myself to ensure this is done correctly. If someone wants to pick it up, or is willing to help me stumble through it, I'd be happy to take a stab at it. Thank you all for your time reviewing these changes, and let me know how you want to proceed.

thecode:
> I am not well versed in any of these technologies, and I wouldn't trust myself to ensure this is done correctly. If someone wants to pick it up, or is willing to help me stumble through it, I'd be happy to take a stab at it. Thank you all for your time reviewing these changes, and let me know how you want to proceed.

Thanks @GarrStau, we plan to test the zeroconf behaviour, just a little bit busy, let's keep the PR open for now and we update here.

chemelli74:
@GarrStau, do you mind rebasing this PR ?
I would like to check all is fine with `gen2` and `gen3` and finally test `gen1`.

I already know that zeroconf packet are not generated for either of the two.
Will ask Shelly to add this feature.

GarrStau:
> @GarrStau, do you mind rebasing this PR ? I would like to check all is fine with `gen2` and `gen3` and finally test `gen1`.
> 
> I already know that zeroconf packet are not generated for either of the two. Will ask Shelly to add this feature.

@chemelli74 I have rebased without testing for now, I will try to gather some devices to test in the next week. Thank you for picking this back up!

GarrStau:
@chemelli74 as far as I can tell the changes are working and stable for a single Gen 2 device on range extender, although I went through the code again and had a question - should aioshelly.rpc_device.wsrpc.WsServer.subscriptions be modified to use the port in addition to the IP address (f"{ip}:{port}") as the dictionary key?
I'm still unsure of the finer details of the Range Extender, or how CoAP works. But I don't see a way to configure a Gen 1 device on Range Extender (and therefore on subnet 192.168.33.0/24) to communicate with my local HA instance.

chemelli74:
So far `gen1` are not supported by Shelly as CoAP port is not forwarded.
They are considering this additional feature but for now there is no ETA or commitment.

GarrStau:
Thoughts on adding the port to the key in `aioshelly.rpc_device.wsrpc.WsServer.subscriptions`? I'm admittedly quite naive how homeassistant core works, but I think this is required to make the WsServer work correctly with IPs. If subscriptions are stored by MAC, this wouldn't be a problem.
Without a deeper understanding of how the port forwarding is implemented, I'm not sure if new issues will crop up with ports changing that would require workarounds in the integration to re-enumerate devices based on the local device's `WiFi.ListAPClients`. I can dig into that, but don't want to spend a bunch of time going in the wrong direction.

GarrStau:
@dolai1 as requested I have rebased. I hope it works for you, but note this PR has been open for over a year already so don't expect a quick resolution (there are other features that I don't have the skill or time to verify). I have not looked at homeassistant core to see how much work is required on that end to allow specifying port in the UI. Let me know if I can be of any help, but at this point it's a rarely used feature asking for a good chunk of dev time.

chemelli74:
Hi @GarrStau, do you mind joining HomeAssistant  on [Discord](https://discord.com/invite/home-assistant) and then contact me: `chemelli74#2180`  ?

aioshelly/block_device/device.py
```diff
@@ -112,6 +112,11 @@ def ip_address(self) -> str:
         """Device ip address."""
         return self.options.ip_address
 
+    @property
+    def port(self) -> int:
+        """Device ip address."""
+        return self.options.port
```
bieniu, line 118: Why do you add the `port` property for gen1 devices if the change only applies to gen2+ devices?

chemelli74: Indeed, a leftover of the first attempt to add also gen1 support but as we know from Shelly it won't be available.

chemelli74: Reverted all changes for `block_device`

chemelli74:
I don't know why pushing a change for resolving the conflict closed the PR.
@GarrStau can you reopen it ?
Sorry for the trouble

aioshelly/ble/__init__.py
```diff
@@ -92,7 +92,7 @@ def create_scanner(
         name,
         HaBluetoothConnector(
             # no active connections to shelly yet
-            client=None,  # type: ignore[arg-type]
```
chemelli74, line 95: Local mypy wants this removed, CI wants it added.

aioshelly/rpc_device/device.py
```diff
@@ -171,7 +181,7 @@ async def initialize(self, async_init: bool = False) -> None:
             self.initialized = True
         except InvalidAuthError as err:
             self._last_error = InvalidAuthError(err)
-            _LOGGER.debug("host %s: error: %r", ip, self._last_error)
+            _LOGGER.debug("host %s:%s: error: %r", ip, self.port, self._last_error)
```
bieniu, line 184: I think we should be consistent here, either we use local variables `ip` and `port` or `self.ip_address` and `self.port` everywhere.

aioshelly/rpc_device/device.py
```diff
@@ -180,13 +190,13 @@ async def initialize(self, async_init: bool = False) -> None:
             self.initialized = True
         except (MacAddressMismatchError, FirmwareUnsupported) as err:
             self._last_error = err
-            _LOGGER.debug("host %s: error: %r", ip, err)
+            _LOGGER.debug("host %s:%s: error: %r", ip, self.port, err)
```
bieniu, line 193: I think we should be consistent here, either we use local variables `ip` and `port` or `self.ip_address` and `self.port` everywhere.

aioshelly/rpc_device/device.py
```diff
@@ -180,13 +190,13 @@ async def initialize(self, async_init: bool = False) -> None:
...
             if not async_init:
                 await self._disconnect_websocket()
                 raise
         except (*CONNECT_ERRORS, RpcCallError) as err:
             self._last_error = DeviceConnectionError(err)
-            _LOGGER.debug("host %s: error: %r", ip, self._last_error)
+            _LOGGER.debug("host %s:%s: error: %r", ip, self.port, self._last_error)
```
bieniu, line 199: I think we should be consistent here, either we use local variables `ip` and `port` or `self.ip_address` and `self.port` everywhere.

bieniu (commented):
How is `aioshelly` protected against gen2+ battery powered devices and gen1 behind the range extender? We know they don't work properly in this configuration.

chemelli74:
> How is `aioshelly` protected against gen2+ battery powered devices and gen1 behind the range extender? We know they don't work properly in this configuration.

`block_device` code in aioshelly is not changed at all.
gen2+ battery devices are currently checked by Shelly as support seems easy to add.

bdraco:
What does the MDNS advertisement look like when it's coming through the range extender?

chemelli74:
> What does the MDNS advertisement look like when it's coming through the range extender?

There is no MDNS unfortunately.
Just manual configuration.

bieniu:
> gen2+ battery devices are currently checked by Shelly as support seems easy to add.

aioshelly accepts a battery-powered device that is behind the range extender and tries to communicate with it. Of course it doesn't receive any status updates from the device. In my opinion, the library should reject the battery-powered device because it is not supported.

obraz



> `block_device` code in aioshelly is not changed at all

If you try to use a gen1 device that is behind the range extender with aioshelly the result will be `WrongShellyGen: Wrong Shelly generation 1, device gen: 2`. This doesn't say anything about the real reason the error occurred. IMO there should be a dedicated exception for such an event.

chemelli74:
Kiril says battery powered should work. I'll test them more later.
About the exception, I fear you missed it in your first review. Let's see what we can do to improve it.

chemelli74:
@bieniu just noticed that you used "--init" on your command line, while it should be omitted for a non connected device like H&T

bieniu:
Yeah

> while it should be omitted for a non connected device like H&T

Yeah my mistake

bdraco:
> > What does the MDNS advertisement look like when it's coming through the range extender?
> 
> There is no MDNS unfortunately. Just manual configuration.

I was worried we would switch the port back but it looks like we don't do that 

Are we going to change the ip to the wrong one from https://github.com/home-assistant/core/blob/fc4b18b907ebccba3b2eaa9e7b4b80165f559e4f/homeassistant/components/shelly/config_flow.py#L238 ?

bdraco:
Screenshot 2024-02-28 at 7 23 52 AM

Discussed on discord with @chemelli74 and he confirmed the range extender doesn't forward mdns packets so my concern is likely not an issue

bdraco (approved):

chemelli74:
Thx!

Issue #182, opened 2022-01-02
Support for proxy mode to support docker containers and reverse proxies


ruimarinho:
Hi,

I'm running Home Assistant inside docker using isolated VLAN networks in bridge mode with CoAP unicast integration. Requests and replies using CoAP work fine because the request is initiated by the docker container, therefore the reply is correctly identified with the right Shelly device.

However, periodic and local push messages are initiated by Shelly devices and consequently the CoAP message has the IP of the Home Assistant container and not the IP of the Shelly device (e.g. a docker internal container IP of `172.21.0.4` instead of `192.168.100.195`, assuming an IoT VLAN of `192.168.100.0/24`).

One of the ideas I've been playing around is to use a CoAP Forwarder/Relayer running on a macvlan on the main subnet which listens to UDP datagrams and adds a tag (`prxy`) with the original IP address (like an `X-Forwarded-For` HTTP header). See the original idea [here](https://github.com/StyraHem/ShellyForHASS/blob/master/util/proxy.py#L26-L28).

The problem with this approach is that aioshelly does not understand that tag, resulting in an error like `Invalid Message from unknown host 172.21.0.8: Wrong type, 114`. I'm unsure if this is part of the CoAP standard or some custom modification, but the issue is real and hard to solve without support for a similar setting.

Here's the output with (first line) and without the tag (second line) appended:

```
coap                | b'P\x1e\x96V\xb3cit\x01s\xed\x0b\xec\x00SHEM#B9EA81#2\xd2C\x96\x00\x82\x19\x00\xff{"G":[[0,9103,0],[0,1101,0],[0,4105,552.74],[0,4106,4067801.3],[0,4107,570811.0],[0,4108,241.90],[0,4205,4.49],[0,4206,1680482.2],[0,4207,0.0],[0,4208,241.90],[0,6102,0]]}'
coap                | bytearray(b'prxy\xc0\xa8d\xc3P\x1e\x96V\xb3cit\x01s\xed\x0b\xec\x00SHEM#B9EA81#2\xd2C\x96\x00\x82\x19\x00\xff{"G":[[0,9103,0],[0,1101,0],[0,4105,552.74],[0,4106,4067801.3],[0,4107,570811.0],[0,4108,241.90],[0,4205,4.49],[0,4206,1680482.2],[0,4207,0.0],[0,4208,241.90],[0,6102,0]]}')
```

Looking forward to hear your thoughts.

thecode:
[...]

PR #327, opened 2023-01-08
Add support for subscribing to updates by device MAC

thecode:
Add options to subscribe for updates using device MAC (device ID)
Subscribing using device ID will help to support devices behind NAT, proxies, or if device IP changes
Implementation is backward compatible and allow subscribing for updates either by MAC or device IP, we can later drop the support for subscribing by IP.

Closes https://github.com/home-assistant-libs/aioshelly/issues/182

Gen1:
- CoAP contains the device ID in CoAP option 3332, for this parsing of options was added, logic for parsing options is based on https://github.com/chrysn/aiocoap/blob/master/aiocoap/options.py
- For compatibility with older Gen1 devices that only include the last 6 chars of MAC, comparison only compare the last 6 MAC chars

Gen2:
- Device ID is sent on every frame as `src`

Checked on multiple Gen1/Gen2 devices with various firmware versions for Gen1.
Update `example.py` to test subscribing by MAC, for testing with debug enabled:
```
python example.py --ip 192.168.1.75 -i -deb -m "ABCDEF123456"
```

### DIFF
--- a/aioshelly/block_device/coap.py
+++ b/aioshelly/block_device/coap.py
@@ -2,12 +2,13 @@
 from __future__ import annotations
 
 import asyncio
-import json
 import logging
 import socket
 import struct
 from types import TracebackType
 from typing import Callable, cast
+
+from ..json import JSONDecodeError, json_loads
 
 COAP_OPTION_DEVICE_ID = 3332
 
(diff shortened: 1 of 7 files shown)


aioshelly/block_device/coap.py
```diff
@@ -36,9 +39,34 @@ def __init__(self, sender_addr: tuple[str, int], payload: bytes) -> None:
...
+            self.options[option_number] = raw_data[:length]
+            raw_data = raw_data[length:]
+
+        if not data:
+            raise InvalidMessage("Received message without data")
+
         try:
-            self.payload = json.loads(payload.rsplit(b"\xff", 1)[1].decode())
-        except (json.decoder.JSONDecodeError, UnicodeDecodeError, IndexError) as err:
+            self.payload = json.loads(data.decode())
```
bdraco, line 68:
```suggestion
            self.payload = json_loads(data.decode())
```

Also needs
`from ..json import json_dumps, json_loads`

bdraco:
All my Shelly Plus Plug US work fine with this change

bieniu (approved):
1L and Plug S work fine, looks good :+1:

thecode:
Thanks @bieniu @bdraco

PR #336, opened 2023-01-20
Add name property and cleanup properties

thecode:
- Add name property to have common API for Gen1 & Gen2
- Cleanup properties
- Update example to show device name


frenck (approved):

thecode:
Thanks!

PR #339, opened 2023-01-23
Bump version to 5.3.0

thecode:
Mainly to include https://github.com/home-assistant-libs/aioshelly/pull/332 since these devices start shipping and we don't have thier model names in `5.2.1`


bdraco (approved):

thecode:
Thanks!

PR #376, opened 2023-04-13
Add an extra line for Shelly Plus 2PM (new model name)

emilbratt:
My new Shelly Plus 2PM`s model name is not recognised.
I found that there exist an entry for my model, but with model name SNSW-002P16EU.
My new shelly device has the model name SNSW-102P16EU.

Below is some of the output from running example.py
```
** shelly_alias - Unknown (SNSW-102P16EU)  @ 192.168.1.10 **
```

I added a 2nd entry to const.py with the new model name and a trailing comment.

### DIFF
--- a/aioshelly/const.py
+++ b/aioshelly/const.py
@@ -70,6 +70,7 @@ MODEL_NAMES = {
     "SNSW-001X16EU": "Shelly Plus 1",
     "SNSW-0024X": "Shelly Plus I4",
     "SNSW-002P16EU": "Shelly Plus 2PM",
+    "SNSW-102P16EU": "Shelly Plus 2PM",  # hw v2
     "SPEM-003CEBEU": "Shelly Pro 3EM",
     "SPSW-001PE16EU": "Shelly Pro 1PM",
     "SPSW-001XE16EU": "Shelly Pro 1",


aioshelly/const.py
```diff
@@ -70,6 +70,7 @@
     "SNSW-001X16EU": "Shelly Plus 1",
     "SNSW-0024X": "Shelly Plus I4",
     "SNSW-002P16EU": "Shelly Plus 2PM",
+    "SNSW-102P16EU": "Shelly Plus 2PM",  # hw v2
```
bieniu, line 73: Where does the information that this is version 2 hardware come from?

emilbratt: > Where does the information that this is version 2 hardware come from?

I took inspiration from the "Shelly Button1" where I found 2 entries for the same product.
```
    "SHBTN-1": "Shelly Button1",
    "SHBTN-2": "Shelly Button1",  # hw v2
```
Then I replicated it without doing any re-search and gave no more thoughts to this.
I understand now that this was not the right idea.

The Shelly Plus 2PM device I added was bought in Norway if that is of any help.
I can try to gather more info about the device and find out why there exists an extra model name for it.

bieniu: We are waiting for an answer from Shelly dev team whether it is really a new hardware version.

emilbratt: Great!
Don't know if it is of any help, but here is the JSON output from my shelly plus 2pm device when requesting the **http://<host>/shelly** endpoint.
```
name: "shelly_alias"
id: "shellyplus2pm-441793cf094c"
mac: "441793CF094C"
model: "SNSW-102P16EU"
gen: 2
fw_id: "20221206-141957/0.12.0-gafc2404"
ver: "0.12.0"
app: "Plus2PM"
auth_en: true
auth_domain: "shellyplus2pm-441793cf094c"
profile: "switch"
```

thecode (changes requested):

bieniu (approved):
LGTM, thanks @emilbratt

thecode (approved):
Thanks!

PR #392, opened 2023-06-13
Speed up parsing bluetooth advertisement data

bdraco:
We can use `a2b_base64` since we know the output is always going to be ascii


bieniu (approved):
Looks good and works without problems 👍

bdraco:
thanks

PR #405, opened 2023-07-27
Gather asyncio tasks in a collection

bieniu:
Related to:

obraz

### DIFF
--- a/aioshelly/rpc_device/wsrpc.py
+++ b/aioshelly/rpc_device/wsrpc.py
@@ -8,6 +8,7 @@ import logging
 import socket
 import time
 from asyncio import Task, tasks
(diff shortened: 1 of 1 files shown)


aioshelly/rpc_device/wsrpc.py
```diff
@@ -191,7 +193,7 @@ async def connect(self, aiohttp_session: aiohttp.ClientSession) -> None:
                 err,
             )
 
-        self._rx_task = asyncio.create_task(self._rx_msgs())
+        self._rx_task = self._create_and_track_task(self._rx_msgs())
```
bdraco, line 196: this one can stay since `self._rx_task` is a strong reference

PR #408, opened 2023-08-05
suppress update callbacks when nothing changes

bdraco:
avoid making HA write state when the data is the same

we need to filter out `'time': '19:04', 'unixtime': 1691211882, 'uptime': 5232950,`


bdraco:
The callbacks are legit, but its mostly power sensors outside the margin of error 😢

bdraco:
The big issue here is we write state on all entities when only the power sensor changes

thecode:
> The big issue here is we write state on all entities when only the power sensor changes

Maybe we can introduce delta updates instead of merging the status

PR #414, opened 2023-08-14
Drop Python 3.9 support

thecode:
# Drop Python 3.9 support

minor fix for pylintrc
```
pylintrc:1:0: R0022: Useless option value for '--disable', 'abstract-class-little-used' was removed from pylint, see https://pylint.pycqa.org/en/latest/whatsnew/1/1.4.html#what-s-new-in-pylint-1-4-3. (useless-option-value)
```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,7 +15,6 @@ jobs:
     strategy:
       matrix:
         python-version:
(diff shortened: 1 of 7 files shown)


frenck (approved):

thecode:
Thanks!

PR #415, opened 2023-08-18
Abort if input MAC does not match the device MAC

thecode:
Abort connection to device and raise `MacAddressMismatchError` if the supplied MAC address doesn't match the MAC address returned from the device in `/shelly`.

Since we supply the MAC address to allow routing incoming messages we can detect if the IP is wrong and we are connecting to another Shelly device.

I had few IPs mixed up after power failure 
1. Shelly IP mixed with a non Shelly device --> This is not harmful, Init will fail until HA detect the new IP from the device (DHCP/mDNS).
2. Shelly Gen1 IP mixed with Shelly Gen2 IP --> This will raise `WrongShellyGen` which is not handled by HA since it shouldn't happen (runtime error).
3. Shelly IP Mixed with another Shelly from the same Gen --> This creates a bad situation since init succeeds and entities from the one Shelly are added to the other Shelly and HA controls the wrong device.

By aborting we allow the init to fail until we get a new IP by DHCP or mDNS.

To test with wrong MAC:
```
python example.py --ip 192.168.1.49 --mac 1234A2E0BBA5 -i
MAC address mismatch, error: MacAddressMismatchError('Input MAC: 1234A2E0BBA5, Shelly MAC: 1234A2E0BBA6')
```

### DIFF
--- a/aioshelly/block_device/device.py
+++ b/aioshelly/block_device/device.py
@@ -18,6 +18,7 @@ from ..const import CONNECT_ERRORS, DEVICE_IO_TIMEOUT, HTTP_CALL_TIMEOUT
 from ..exceptions import (
     DeviceConnectionError,
     InvalidAuthError,
+    MacAddressMismatchError,
     NotInitialized,
     ShellyError,
     WrongShellyGen,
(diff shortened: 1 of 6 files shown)


bieniu (approved):
Looks good, I tested this change with gen1 and gen2 devices, works as intended, thanks @thecode

PR #426, opened 2023-09-09
Create futures with loop.create_future()

bdraco:
https://docs.python.org/3/library/asyncio-future.html#asyncio.Future

> The rule of thumb is to never expose Future objects in user-facing APIs, and the recommended way to create a Future object is to call loop.create_future(). This way alternative event loop implementations can inject their own optimized implementations of a Future object.

### DIFF
--- a/aioshelly/rpc_device/wsrpc.py
+++ b/aioshelly/rpc_device/wsrpc.py
@@ -113,7 +113,7 @@ class RPCCall:
 
     __slots__ = ("auth", "call_id", "params", "method", "src", "dst", "resolve")
 
-    def __init__(
+    def __init__(  # pylint: disable=too-many-arguments
         self,
         call_id: int,
         method: str,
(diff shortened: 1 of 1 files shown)


aioshelly/rpc_device/wsrpc.py
```diff
@@ -394,13 +397,14 @@ async def _rpc_call(
...
         self._calls[call.call_id] = call
 
         try:
             async with async_timeout.timeout(timeout):
                 await self._send_json(call.request_frame)
-                resp: dict[str, Any] = await call.resolve
+                resp = await future
         except asyncio.TimeoutError as exc:
             with contextlib.suppress(asyncio.CancelledError):
                 call.resolve.cancel()
```
MartinHjelmare, line 410: Should this also be changed to cancel and await `future`?

bdraco: It still works either way, but cleaner for sure

bdraco: will adjust shortly

bdraco: d834485f833336137d86c4625b495742a4af5752

MartinHjelmare (approved):

bdraco:
thanks

PR #444, opened 2023-11-06
Add support for devices with multiple profiles

monester:
For example ShellyPlus 2PM can be configured as switch with 2 relays or a cover.

`curl http://192.168.1.133/rpc/Shelly.ListProfiles`

```json
{
  "profiles": {
    "cover": {
      "components": [
        {
          "type": "input",
          "count": 2
        },
        {
          "type": "cover",
          "count": 1
        }
      ]
    },
    "switch": {
      "components": [
        {
          "type": "input",
          "count": 2
        },
        {
          "type": "switch",
          "count": 2
        }
      ]
    }
  }
}
```

### DIFF
--- a/aioshelly/rpc_device/device.py
+++ b/aioshelly/rpc_device/device.py
@@ -165,7 +165,6 @@ class RpcDevice:
             async with async_timeout.timeout(DEVICE_IO_TIMEOUT):
                 await self._wsrpc.connect(self.aiohttp_session)
                 await self.update_config()
-                await self.update_device_info()
                 await self.update_profiles()
 
                 if not async_init or self._status is None:
@@ -238,21 +237,26 @@ class RpcDevice:
         """Get device config from 'Shelly.GetConfig'."""
         self._config = await self.call_rpc("Shelly.GetConfig")
 
-    async def update_device_info(self) -> None:
-        """Get device config from 'Shelly.GetDeviceInfo'."""
-        self._device_info = await self.call_rpc("Shelly.GetDeviceInfo")
-
     async def update_profiles(self) -> None:
         """Get device config from 'Shelly.ListProfiles'."""
-        if self._device_info and "profile" in self._device_info:
+        if "profile" in self.shelly:
             self._profiles = (await self.call_rpc("Shelly.ListProfiles"))["profiles"]
 
     async def set_profile(self, profile_name: str) -> None:
         """Set device profile via 'Shelly.SetProfile'."""
-        if self._profiles and profile_name in self._profiles:
+        if self._shelly.get('profile') and self._shelly['profile'] != profile_name:
             await self.call_rpc("Shelly.SetProfile", {"name": profile_name})
-        else:
-            raise ValueError("Invalid profile_name")
+            await self.shutdown()
+            # device is rebooted on profile_name change as profile is changed
+            # 3 seconds sleep is needed to wait for device to perform reboot action
+            # as there is no way to determine is device rebooted
+            await asyncio.sleep(2)
+            while self._shelly['profile'] != profile_name:
+                await asyncio.sleep(1)
+                self._shelly = await get_info(
+                    self.aiohttp_session, self.options.ip_address, self.options.device_mac
+                )
+            await self.initialize()
 
     async def script_list(self) -> list[ShellyScript]:
         """Get a list of scripts from 'Script.List'."""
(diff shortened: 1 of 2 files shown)


bieniu:
Please explain the context of this change. Are you going to implement the profile changing in HA Shelly integration?

monester:
Context is following - I have shellyplus 2pm connected to the roll-up cover. I tried to configure it via HA but didn't understand why it was showing 2 switches instead of cover ui. Then I found out that it is a dedicated "Device profile" setting in the Shelly app so I've changed profile there and now it is working properly in HA.

But I miss this functionality in HA. So I want to add this option to the shelly integration. I have a prototype already working so I'm testing now adding new device or changing mode on existing one.

bieniu:
The device operating mode is set in the device configuration. HA integration adapts to the selected device operating mode. Honestly, I'm not sure if we should allow changing the operating mode of an HA device.

monester:
Why not to allow configuration of device? Especially when new device is added. It is exactly my case - I've connected with shelly app to my wifi and then added to the HA. And then I was trying to figure out the whole evening why it doesn't work as expected.

bieniu:
IMO the workflow should be different, first we configure the device correctly and only then integrate it with HA.

My concern that sooner or later the user will change the operating mode of an already configured device by accident. This completely changes the way the device is configured in HA and can cause a bit of chaos.

Let's wait for feedback from other code owners.

thecode:
I am not against supporting this from HA, however the owner needs to realize this is not simple. If we add support for "on the fly" profile switching the existing entities needs to be cleaned up and reload the device to create the new entities.
I tend to agree with @bieniu that this is not the desired way to configure a device, but if we decide to continue I would like to see a parallel PR in HA core to see the usage of this API

The current implementation suggested here try to pull profile from all Gen2 devices, while only small number of devices actually support profiles (as far as I understand only Shelly 2PM, Shelly Pro 2PM and Shelly Pro 4PM so I don't think we should fetch profiles for every device. Fetching profile for devices that doesn't have profiles fails so the code either needs to handle this differently from other devices or catch the failure. A better approach will be to add the API support for that but don't automatically fetch the profiles and let HA fetch profiles if the device supports profiles, since HA will need to handle this anyhow to cleanup entities so it can't be handled only in `aioshelly`.

aioshelly/rpc_device/device.py
```diff
@@ -233,6 +238,22 @@ async def update_config(self) -> None:
         """Get device config from 'Shelly.GetConfig'."""
         self._config = await self.call_rpc("Shelly.GetConfig")
 
+    async def update_device_info(self) -> None:
+        """Get device config from 'Shelly.GetDeviceInfo'."""
+        self._device_info = await self.call_rpc("Shelly.GetDeviceInfo")
+
```
thecode, line 244: The data here is identical to `/shelly` endpoint which we already fetch from the device. We use the `/shelly` since it is a generic way that doesn't need auth and works on all Shelly devices.

aioshelly/rpc_device/device.py
```diff
@@ -162,6 +165,8 @@ async def initialize(self, async_init: bool = False) -> None:
             async with async_timeout.timeout(DEVICE_IO_TIMEOUT):
                 await self._wsrpc.connect(self.aiohttp_session)
                 await self.update_config()
+                await self.update_device_info()
+                await self.update_profiles()
```
thecode, line 169: As noted in my previous comment, I prefer the user of the lib to do it when needed, by calling `get_profiles` and not storing the data here. This is not supported for most of the devices and sleeping devices might have problems with this additional call.

bieniu:
On second thought... this could be an interesting feature if implemented as part of the device configuration panel rather than as an entity on HA side.

image

monester:
Added functionality that it is possible configure profile when device is added. https://github.com/monester/home-assistant-core/tree/shelly-profiles

aioshelly/rpc_device/device.py
monester: make update_profiles to query api only if "profile" is defined in `/shelly`

chemelli74:
Hi @monester are you wishing to complete the job ?
If so, please resolve conflicts and update code if necessary so we can check and test.

monester:
Don't count on me for now, I'll be back on project when I have more time.

chemelli74:
> Don't count on me for now, I'll be back on project when I have more time.

Ok, let's close this PR then.
Please open a new one when you are ready.

Thx in advance for your contribution.

PR #448, opened 2023-11-20
Add Shelly Blu Gateway model

chemelli74:


bieniu (approved):
Looks good 👍

thecode (commented):
Until now we added devices while keeping the models with lexicographical sort, please fix this in a new PR. Thanks

bieniu:
I'm working on the change related on model consts.

chemelli74:
> Until now we added devices while keeping the models with lexicographical sort, please fix this in a new PR. Thanks

We are thinking to sort by description and not by model.

thecode:
> > Until now we added devices while keeping the models with lexicographical sort, please fix this in a new PR. Thanks
> 
> We are thinking to sort by description and not by model.

When you do that sort all of them, for now it is is sorted by model. We should keep this consistent until this is changed!

bieniu:
https://github.com/home-assistant-libs/aioshelly/pull/449

PR #449, opened 2023-11-20
Add model constants

bieniu:
With this change we will be able to import and use constants in the integration. Changing the sorting of models improves readability.

- changed `Blu Gateway` to `BLU Gateway` according to this: obraz
- added `SNDM-00100WW` (0-10V Dimmer)
- added `SPEM-003CEBEU400` (Pro 3EM-400)


chemelli74 (approved):
LGTM!

PR #451, opened 2023-11-21
Fix devcontainer setup

chemelli74:
Readd pip install env

cc: @MartinHjelmare


MartinHjelmare (approved):
Thanks!

I've tested it locally both with and without the devcontainer.

PR #456, opened 2023-11-29
Handle NotifyStatus frames with id

chemelli74:
Shelly Wall Display has an **`id`** on the NotifyStatus frame:

```
{'id': 0, 'src': 'ShellyWallDisplay-00082216C97B', 'dst': 'user_1', 'method': 'NotifyStatus', 'params': {'ts': 1701283262, 'thermostat:0': {'id': 0, 'enable': False, 'target_C': 21, 'current_C': 24.6, 'output': False, 'schedules': {'enable': False}}}}
```

```
{'id': 1701197400, 'src': 'ShellyWallDisplay-00082216C97B', 'dst': 'user_1', 'method': 'NotifyStatus', 'params': {'temperature:0': {'id': 0, 'tC': 22.8, 'tF': 73.1}, 'humidity:0': {'id': 0, 'rh': 43}, 'illuminance:0': {'id': 0, 'lux': 5, 'illumination': 'dark'}, 'ts': 1701197400}}
```

This break our logic.

This PR fix it in a general way.

### DIFF
--- a/aioshelly/rpc_device/wsrpc.py
+++ b/aioshelly/rpc_device/wsrpc.py
@@ -284,6 +284,7 @@ class WsRPC:
 
     def handle_frame(self, frame: dict[str, Any]) -> None:
         """Handle RPC frame."""
+        _LOGGER.debug("RPC frame: %s", frame)
         if peer_src := frame.get("src"):
             if self._session.dst is not None and peer_src != self._session.dst:
                 _LOGGER.warning(
@@ -296,7 +297,7 @@ class WsRPC:
         if method := frame.get("method"):
             # peer is invoking a method
             params = frame.get("params")
-            if frame_id:
+            if frame_id and method != "NotifyStatus":
                 # and expects a response
                 _LOGGER.debug("handle call for frame_id: %s", frame_id)
                 self._create_and_track_task(self._handle_call(frame_id))


chemelli74:
> It looks like we try to workaround a Shelly bug here, the ID should indicate a request ID, this is why it is 0 for a notify, it looks like the device fills in the timestamp on the ID (see the end of the line), I would prefer Shelly dev review and explain this before we do a workaround.

I think you are missing discord chat lately ;-)

This is my post on Shelly forum : <https://community.shelly.cloud/topic/558-beta-firmware-125-thermostat/?do=findComment&comment=3623>

The code is only the demonstration for a solution.
For sure, I want to see what they say about it before merging.

kzyapkov:
The presence of `id` in a method call frame implies that a response is expected. The `id` should not be there for notifications. This will be fixed soon with a patch release.

KaHa6uc:
The `id` has been removed from the `NotifyStatus` RPC payloads with WallDisplay firmware version `20231205-101653/1.2.5-a42b4f2c`. Please update

thecode:
> The `id` has been removed from the `NotifyStatus` RPC payloads with WallDisplay firmware version `20231205-101653/1.2.5-a42b4f2c`, thus rendering this PR obsolete. Please update.

Thanks for the update, closing this PR as we agree that this is not something to fix here. We'll update back after testing.

chemelli74:
Thx!
Updated and testing.

bieniu:
Looks good after the update. Thanks.

chemelli74:
Confirmed to work fine, thx for the fix @KaHa6uc

PR #459, opened 2023-12-02
Bump Gen2 minimum firmware version to 1.0.0 (2023-08-03)

bieniu:
Firmware 1.0.0 for gen2 devices is a completely new line of software that fixes many problems and introduces many new functionalities. Gen3 devices also use this line of software.

Changelog: https://shelly-api-docs.shelly.cloud/gen2/changelog#100-2023-08-03

Firmware 1.0.0 functionalities related to HA integration:
- Network frequency in status of EM, Switch, Cover components
- New settings for switches - "External consumption type"
- Sys Drop wakeup_period attribute from configuration (deprecated since 0.11.0)
We will be able to remove this method https://github.com/home-assistant/core/blob/c8bb72935d935680ea7a9c5f8bde552f2d8f7dcd/homeassistant/components/shelly/utils.py#L279-L281

### DIFF
--- a/aioshelly/const.py
+++ b/aioshelly/const.py
@@ -195,8 +195,8 @@ HTTP_CALL_TIMEOUT = 10
 # Firmware 1.8.0 release date (CoAP v2)
 GEN1_MIN_FIRMWARE_DATE = 20200812
 
-# Firmware 0.8.1 release date
-GEN2_MIN_FIRMWARE_DATE = 20210921
+# Firmware 1.0.0 release date
+GEN2_MIN_FIRMWARE_DATE = 20230803
 
 WS_HEARTBEAT = 55
 


bieniu:
Should we mark this as breaking change? 🤔

thecode:
> Should we mark this as breaking change? 🤔

This is definitely a breaking change, we should also explain in the PR why we enforce the minimum (incompatible older firmware as far as I remember)

chemelli74 (approved):

thecode (approved):
Thanks for updating the description

bieniu:
Sorry for writing here, but I guess you Shay haven't been active on Discord lately.
I'm wondering how to split HA changes to make them easier to test and review. My idea is to first open PR to bump aioshelly version, which would include increasing the minimum firmware version. This PR would be a breaking change in HA. The second step would be PR adding support for gen3 in HA. What do you think, guys?

PR #460, opened 2023-12-02
Bump Gen1 minimum firmware version to 1.11.0 (2021-07-15)

bieniu:
Due to many fixes and new functionalities introduced in subsequent firmware versions, the minimum firmware version supported by `aioshelly` is increased to 1.11.0 (2021-07-15). This release included the last major change in API functionality (`transition` for lights).

Changelog: https://shelly-api-docs.shelly.cloud/gen1/#changelog

### DIFF
--- a/aioshelly/const.py
+++ b/aioshelly/const.py
@@ -192,8 +192,8 @@ DEVICE_IO_TIMEOUT = 10
 # Timeout used for HTTP calls
 HTTP_CALL_TIMEOUT = 10
 
-# Firmware 1.8.0 release date (CoAP v2)
-GEN1_MIN_FIRMWARE_DATE = 20200812
+# Firmware 1.13.0 release date (this also means 2.1.9 for Motion/Motion 2/TRV)
+GEN1_MIN_FIRMWARE_DATE = 20230503
 
 # Firmware 0.8.1 release date
 GEN2_MIN_FIRMWARE_DATE = 20210921
--- a/aioshelly/const.py
+++ b/aioshelly/const.py
@@ -192,8 +192,8 @@ DEVICE_IO_TIMEOUT = 10
 # Timeout used for HTTP calls
 HTTP_CALL_TIMEOUT = 10
 
-# Firmware 1.13.0 release date (this also means 2.1.9 for Motion/Motion 2/TRV)
-GEN1_MIN_FIRMWARE_DATE = 20230503
+# Firmware 1.11.0 release date
+GEN1_MIN_FIRMWARE_DATE = 20210715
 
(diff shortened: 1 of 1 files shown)


bieniu:
Same question, should we mark this as breaking change?

thecode:
> OK, maybe I was too strict with the version number.
> 
> Therefore I suggest using version 1.11.0 (2021-07-15 according to the [documentation](https://shelly-api-docs.shelly.cloud/gen1/#changelog)) as the minimum version for gen1, this version introduced `transition` for lights.
> 
Sounds great

> I don't remember why in the code the date for 1.11.0 is 2023-02-26 (maybe this was a beta?) https://github.com/home-assistant/core/blob/c8bb72935d935680ea7a9c5f8bde552f2d8f7dcd/homeassistant/components/shelly/const.py#L36-L37
> 
> What do you think?

Looking at the timeline of the releases I assume it was a pre-release firmware we tested. I downgraded a bulb (`SHCB-1`) to `v1.11.0` using [Firmware Archive Link Generator](http://archive.shelly-tools.de/version/v1.11.0/SHCB-1.zip) and it is dated `20210720-184102/v1.11.0-g6abd92e`. I assume the few days difference is due the time it was released to the time it was actually built.

We can use `2021-07-15` as mentioned in the changelog, I think we should also add this in the PR description that the minimum is set this version that supports transitions.

chemelli74:
> > OK, maybe I was too strict with the version number.
> 
> > 
> 
> > Therefore I suggest using version 1.11.0 (2021-07-15 according to the [documentation](https://shelly-api-docs.shelly.cloud/gen1/#changelog)) as the minimum version for gen1, this version introduced `transition` for lights.
> 
> > 
> 
> Sounds great
> 
> 
> 
> > I don't remember why in the code the date for 1.11.0 is 2023-02-26 (maybe this was a beta?) https://github.com/home-assistant/core/blob/c8bb72935d935680ea7a9c5f8bde552f2d8f7dcd/homeassistant/components/shelly/const.py#L36-L37
> 
> > 
> 
> > What do you think?
> 
> 
> 
> Looking at the timeline of the releases I assume it was a pre-release firmware we tested. I downgraded a bulb (`SHCB-1`) to `v1.11.0` using [Firmware Archive Link Generator](http://archive.shelly-tools.de/version/v1.11.0/SHCB-1.zip) and it is dated `20210720-184102/v1.11.0-g6abd92e`. I assume the few days difference is due the time it was released to the time it was actually built.
> 

I got lost, as code there was a reference to 2023-02-06 and one to 2021-07-20; how is just a few days ?

thecode (approved):
Thanks

PR #473, opened 2023-12-30
Fix concurrent call to receive

bdraco:
https://github.com/home-assistant/core/issues/105273

### DIFF
--- a/aioshelly/rpc_device/wsrpc.py
+++ b/aioshelly/rpc_device/wsrpc.py
@@ -11,7 +11,6 @@ from asyncio import Task, tasks
 from collections.abc import Callable, Coroutine
 from dataclasses import dataclass
 from typing import Any, cast
-from yarl import URL
 
 import aiohttp
 import async_timeout
@@ -24,6 +23,7 @@ from aiohttp.web import (
     WebSocketResponse,
     get,
 )
+from yarl import URL
 
 from ..const import NOTIFY_WS_CLOSED, WS_API_URL, WS_HEARTBEAT
 from ..exceptions import (
@@ -261,7 +261,11 @@ class WsRPC:
 
     async def disconnect(self) -> None:
         """Disconnect all sessions."""
-        self._rx_task = None
+        if self._rx_task is not None:
+            self._rx_task.cancel()
+            with contextlib.suppress(asyncio.CancelledError):
+                await self._rx_task
+            self._rx_task = None
         self._cancel_heatbeat_and_pong_response_cb()
         if self._client is None:
             return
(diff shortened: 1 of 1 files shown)


aioshelly/rpc_device/wsrpc.py
```diff
@@ -323,41 +327,52 @@ def handle_frame(self, frame: dict[str, Any]) -> None:
...
+            _LOGGER.debug(
+                "Websocket client connection from %s closed", self._ip_address
+            )
+            self._cancel_heatbeat_and_pong_response_cb()
+            # Ensure the underlying transport is closed
+            # Remove this after aiohttp 3.9.2 is released
+            # as it leaks before this version (and has for
+            # a long time)
+            with contextlib.suppress(Exception):
+                self._client._response.close()  # pylint: disable=protected-access
```
bdraco, line 364: This is not the issue, but it is a problem since aiohttp < 3.9.2 won't close the transport on failure

aioshelly/rpc_device/wsrpc.py
```diff
@@ -261,7 +261,11 @@ def _pong_not_received(self) -> None:
 
     async def disconnect(self) -> None:
         """Disconnect all sessions."""
-        self._rx_task = None
+        if self._rx_task is not None:
+            self._rx_task.cancel()
+            with contextlib.suppress(asyncio.CancelledError):
+                await self._rx_task
+            self._rx_task = None
```
bdraco, line 268: This is likely the problem as we didn't cancel the task on disconnect. But cancelling it isn't safe unless we wrap it in finally to ensure it still cleans up

aioshelly/rpc_device/wsrpc.py
```diff
@@ -323,41 +327,53 @@ def handle_frame(self, frame: dict[str, Any]) -> None:
...
+                    _LOGGER.debug("recv(%s): %s", self._ip_address, frame)
+                except InvalidMessage as err:
+                    _LOGGER.error(
+                        "Invalid Message from host %s: %s", self._ip_address, err
+                    )
+                except ConnectionClosed:
+                    break
+                except Exception:  # pylint: disable=broad-except
+                    _LOGGER.exception("Unexpected error while receiving message")
+                    raise
```
bdraco, line 351: And finally if we did get a RuntimeError on concurrent receive it would infinite loop because it swallowed the exception and would happen again next loop

aioshelly/rpc_device/wsrpc.py
```diff
@@ -323,41 +327,53 @@ def handle_frame(self, frame: dict[str, Any]) -> None:
...
-        self._cancel_heatbeat_and_pong_response_cb()
-
-        for call_item in self._calls.values():
-            if not call_item.resolve.done():
-                call_item.resolve.set_exception(DeviceConnectionError(call_item))
-        self._calls.clear()
-
-        if not self._client.closed:
-            await self._client.close()
+                await self._client.close()
```
bdraco, line 373: Maybe it should do the close last to minimize the risk of a race

bdraco: I'm going to move the close to the end of this block to make sure we clean up before yielding to the event loop

aioshelly/rpc_device/wsrpc.py
```diff
@@ -323,41 +327,57 @@ def handle_frame(self, frame: dict[str, Any]) -> None:
...
+            self._calls.clear()
+
+            client = self._client
+            self._client = None
+            self._on_notification(NOTIFY_WS_CLOSED)
+
+            # Close last since the await can yield
+            # to the event loop and we want to minimize
+            # race conditions
+            if client.closed:
```
bdraco, line 379:
```suggestion
            if not client.closed:
```

bieniu (approved):
I tested this with HA from dev branch and I don't see any side effects. Unfortunately, I wasn't able to reproduce the problem earlier, so I'm not sure if the change will actually fix the problem.

thecode:
How urgent do you think it is? I am going to update my prod this weekend and will try to reproduce the problem, if you think it is urgent we can merge it

bdraco:
For the users with the issue it seems rather urgent since it going to tie up the event loop in an infinite loop, but also no need to rush it out the door as I'd rather have more testing since the number of people it affects is small even if the impact is large.

thecode:
> For the users with the issue it seems rather urgent since it going to tie up the event loop in an infinite loop, but also no need to rush it out the door as I'd rather have more testing since the number of people it affects is small even if the impact is large.

I have aligned my prod to 2024.1.2, I am going to try to reproduce the problem tomorrow, if not reproduced I will apply this patch and we can try to release new `aioshelly` on monday, I believe it will still be before the next core patch release.

thecode:
I tried to reproduce this race (before this patch) but couldn't reproduce it, tried to lock devices to far APs, reboot the APs, disconnect devices, without any errors so I can't help reproducing.
Changes looks good, I suggest we go forward with this change and get feedback from the affected users.
I am updating my prod with this patch so we have more testing.
Thanks @bdraco

thecode (approved):

bieniu:
This change has also been working on my prod since yesterday and I don't see any negative effects.

bdraco:
Its been running on all my production systems for a while will no ill effects.  Let's merge it now

bdraco:
Thanks for the 👀

PR #479, opened 2024-01-10
Fix gen3 names

bieniu:
According to the knowledge base, these models have different names https://kb.shelly.cloud/knowledge-base/shelly-mini-gen3-devices


aioshelly/const.py
```diff
@@ -100,9 +100,9 @@
 MODEL_PRO_EM3_400 = "SPEM-003CEBEU400"
 MODEL_WALL_DISPLAY = "SAWD-0A1XX10EU1"
 # Gen3 RPC based models
-MODEL_PLUS_1_MINI_G3 = "S3SW-001X8EU"
-MODEL_PLUS_1PM_MINI_G3 = "S3SW-001P8EU"
-MODEL_PLUS_PM_MINI_G3 = "S3PM-001PCEU16"
+MODEL_1_MINI_G3 = "S3SW-001X8EU"
+MODEL_1PM_MINI_G3 = "S3SW-001P8EU"
+MODEL_PM_MINI_G3 = "S3PM-001PCEU16"
```
bieniu, line 105: These constants have not been used in the integration code yet, so I think we can rename them. If you disagree, I will restore the old names.

chemelli74: I think the rename is safe

thecode: I don't see any problem to rename them

chemelli74 (approved):

PR #484, opened 2024-01-15
Get firmware info from Shelly update server

chemelli74:
Retrieve the firmware version ( es. "20231219-133948/1.1.0-g34b5d4f") from the official URLs verified with Shelly.

This info will be used only to inform the users that a new version is available.
Code doesn't retrieve any URL so no action can be performed on the device.

### DIFF
--- a/aioshelly/rpc_device/device.py
+++ b/aioshelly/rpc_device/device.py
@@ -7,11 +7,17 @@ from collections.abc import Callable
 from enum import Enum, auto
 from typing import Any, cast
 
-import aiohttp
 import async_timeout
-from aiohttp.client import ClientSession
+from aiohttp import ClientSession
+from yarl import URL
 
-from ..common import ConnectionOptions, IpOrOptionsType, get_info, process_ip_or_options
+from ..common import (
+    ConnectionOptions,
+    IpOrOptionsType,
+    get_firmware_from_web,
+    get_info,
+    process_ip_or_options,
+)
 from ..const import CONNECT_ERRORS, DEVICE_IO_TIMEOUT, NOTIFY_WS_CLOSED
 from ..exceptions import (
     DeviceConnectionError,
@@ -62,7 +68,7 @@ class RpcDevice:
     def __init__(
         self,
         ws_context: WsServer,
-        aiohttp_session: aiohttp.ClientSession,
+        aiohttp_session: ClientSession,
         options: ConnectionOptions,
     ):
         """Device init."""
@@ -72,6 +78,7 @@ class RpcDevice:
         self._status: dict[str, Any] | None = None
         self._event: dict[str, Any] | None = None
         self._config: dict[str, Any] | None = None
+        self._info: dict[str, Any] | None = None
         self._wsrpc = WsRPC(options.ip_address, self._on_notification)
         sub_id = options.ip_address
         if options.device_mac:
@@ -87,7 +94,7 @@ class RpcDevice:
     @classmethod
     async def create(
         cls,
-        aiohttp_session: aiohttp.ClientSession,
+        aiohttp_session: ClientSession,
         ws_context: WsServer,
         ip_or_options: IpOrOptionsType,
         initialize: bool = True,
@@ -163,6 +170,7 @@ class RpcDevice:
             async with async_timeout.timeout(DEVICE_IO_TIMEOUT):
                 await self._wsrpc.connect(self.aiohttp_session)
                 await self.update_config()
+                await self.update_info()
 
                 if not async_init or self._status is None:
                     await self.update_status()
@@ -233,6 +241,10 @@ class RpcDevice:
     async def update_config(self) -> None:
         """Get device config from 'Shelly.GetConfig'."""
         self._config = await self.call_rpc("Shelly.GetConfig")
+
+    async def update_info(self) -> None:
+        """Get device status from 'Shelly.GetDeviceInfo'."""
+        self._info = await self.call_rpc("Shelly.GetDeviceInfo")
 
     async def script_list(self) -> list[ShellyScript]:
         """Get a list of scripts from 'Script.List'."""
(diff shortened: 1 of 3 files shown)


aioshelly/rpc_device/device.py
```diff
@@ -233,6 +241,10 @@ async def update_config(self) -> None:
         """Get device config from 'Shelly.GetConfig'."""
         self._config = await self.call_rpc("Shelly.GetConfig")
 
+    async def update_info(self) -> None:
+        """Get device status from 'Shelly.GetDeviceInfo'."""
+        self._info = await self.call_rpc("Shelly.GetDeviceInfo")
+
```
thecode, line 247: The output of this endpoint is identical to `/shelly` and we already fetch at  https://github.com/home-assistant-libs/aioshelly/blob/1f0b69a1541b32b5c9a03752f46f1b9fc9573883/aioshelly/rpc_device/device.py#L155
We used `/shelly` since older devices didn't expose the `Shelly.GetDeviceInfo` (or at least without authentication and this allows us to have an identical path for gen1/2/3

chemelli74: Unfortunately ".sys.app" is only present on the `DeviceInfo`. Tested on 5 different models :-(

chemelli74: BTW, we need the "app" model only for gen2, gen1 provides a single URL for all devices.

thecode: I checked few devices and they all have the same output which also includes the `app` would be better to first ask Shelly why you see a difference on your devices

chemelli74: Let me recheck,if it's /shelly is a lot better

chemelli74: Double checked on `WallDisplay` and on `BLU Gateway`, both missing.

thecode (changes requested):
I am against adding Shelly cloud access until we finalize the discussion around it.
There was also a concern from Shelly how we are going to make sure we cache the response and don't do it too frequently which is not addressed in this PR.

aioshelly/rpc_device/device.py
kzyapkov: WallDisplay is an Android device, it impelents the next-gen apis but is based on a different platform and does not shrare code. `app` may not be there.

BLU Gateway does have the `app` property in `/shelly`

kzyapkov:
I am also against having this functionality in aioshelly. Albeit rare, devices may require specific upgrade paths for some older versions. If the wrong firmware is flashed on a device it may brick. Our update server gets identifying bits from the devices when they ask for new versions and takes the necessary steps for flashing intermediate versions.

We are discussing adding a public repository with stable and previous stable releases, but what you have now is not it.

chemelli74:
> Albeit rare, devices may require specific upgrade paths for some older versions. If the wrong firmware is flashed on a device it may brick. Our update server gets identifying bits from the devices when they ask for new versions and takes the necessary steps for flashing intermediate versions.

@kzyapkov , this PR is not about flashing the firmware, but only retrieve the information of the latest available (thus the title).
Is in charge to the user to flash it.

On the other hand, take care that we have a update firmware button in HA that will flash latest.
This functionality is there since ages. So far, AFAIK, we didn't had a single report of a issue.

> We are discussing adding a public repository with stable and previous stable releases, but what you have now is not it.

Nice! Thx.

thecode:
> On the other hand, take care that we have a update firmware button in HA that will flash latest.
> This functionality is there since ages. So far, AFAIK, we didn't had a single report of a issue.

As I already tried to explain on discord (which you try to ignore) when we ask the device to trigger an update we are not providing the firmware URL. We just trigger update to latest which is  the same as the update button on the device web, the firmware URL is reported to the device based on multiple factors as @kzyapkov explained above.

I still insist we shouldn't provide any URL to the user, a simple example is Shelly RGBW2 which if uses the URL provided by this design will flash an incorrect firmware to the device (I can explain you the specific logic for this devices, but I don't know what is the logic in all devices), this is just one example and I am not sure how this will work for the entire install base and if there are other devices that can break.

kzyapkov:
You could let devices fetch update info -- call [Shelly.CheckForUpdate](https://shelly-api-docs.shelly.cloud/gen2/ComponentsAndServices/Shelly#shellycheckforupdate)

chemelli74:
@thecode, please check the code.
It DOESN'T fetch any url at all.
It just retrieve the firmware build string, nothing else.

chemelli74:
> You could let devices fetch update info -- call [Shelly.CheckForUpdate](https://shelly-api-docs.shelly.cloud/gen2/ComponentsAndServices/Shelly#shellycheckforupdate)

The issue is linked to misconfigured devices that don't have internet access and don't notify the user of a new available firmware.
The idea is just to make them aware a new version exist.
But no action at all will ever be performed.

chemelli74:
@thecode, @kzyapkov, sorry for the imprecision: I just noticed that I wrote "button" instead of "service".

What I wanted to focus is that we have a service in HA that can be used to flash device providing a URL to the OTA endpoint.
If this can brick the device, we should that avoid the possibility to use custom URLs.

aioshelly/rpc_device/device.py
chemelli74: Wall Display was on a old firmware, latest works fine.

BLU Gateway needed a hard reset, now works fine.

bieniu:
> What I wanted to focus is that we have a service in HA that can be used to flash device providing a URL to the OTA endpoint.

There is `url` parameter in `aioshelly` for gen1 devices but we don't use it in the integration. So we only run the API methods to start an OTA update.

https://github.com/home-assistant/core/blob/83e0a7528d8052005c338530ceb430e52e3ce08a/homeassistant/components/shelly/update.py#L212


https://github.com/home-assistant/core/blob/83e0a7528d8052005c338530ceb430e52e3ce08a/homeassistant/components/shelly/update.py#L297

chemelli74:
> > What I wanted to focus is that we have a service in HA that can be used to flash device providing a URL to the OTA endpoint.
> 
> There is `url` parameter in `aioshelly` for gen1 devices but we don't use it in the integration. So we only run the API methods to start an OTA update.

Yes, if you look at the trigger_ota_update(), but the service itself allows the url parameter in its call.
This means that a user can specify a custom url for update. 

As this can in rare cases be a issue as @kzyapkov wrote, I just wanted to point it out.

bieniu:
> but the service itself allows the url parameter in its call

Which service?

chemelli74:
I still think this simple notification (no action at all) could be very useful for all users that are not notified of a new firmware for whatever reason.

But Team decided not to go this way, so I close this PR.

PR #485, opened 2024-01-18
Trigger OTA for not initialized devices

chemelli74:
Allow OTA to work for devices that are not initialized.
This will be used in HA for repairing an unsupported firmware for connected devices.

### DIFF
--- a/aioshelly/common.py
+++ b/aioshelly/common.py
@@ -153,6 +153,7 @@ async def trigger_ota_http(
     beta: bool = False,
 ) -> bool:
     """Trigger a firmware update via OTA http endpoint."""
+    auth: aiohttp.BasicAuth | dict[str, Any] | None = None
     if gen in BLOCK_GENERATIONS:
         path = "/ota"
         auth = aiohttp.BasicAuth(username, password) if username and password else None
--- a/aioshelly/common.py
+++ b/aioshelly/common.py
@@ -13,6 +13,7 @@ import aiohttp
 from yarl import URL
 
 from .const import (
+    BLOCK_GENERATIONS,
     CONNECT_ERRORS,
     DEVICE_IO_TIMEOUT,
     GEN1_LIGHT_TRANSITION_MIN_FIRMWARE_DATE,
@@ -20,6 +21,7 @@ from .const import (
     GEN1_MODELS_SUPPORTING_LIGHT_TRANSITION,
     GEN2_MIN_FIRMWARE_DATE,
     GEN3_MIN_FIRMWARE_DATE,
+    HTTP_CALL_TIMEOUT,
 )
 from .exceptions import (
     DeviceConnectionError,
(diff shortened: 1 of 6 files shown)


aioshelly/common.py
```diff
@@ -137,3 +139,23 @@ def shelly_supported_firmware(result: dict[str, Any]) -> bool:
...
+
+
+async def trigger_ota_http(session: aiohttp.ClientSession, host: str, gen: int) -> bool:
+    """Trigger a firmware update via OTA http endpoint."""
+    if gen in BLOCK_GENERATIONS:
+        path = "ota"
+    else:
+        path = "rpc/Shelly.Update"
+    try:
+        await session.get(
```
bieniu, line 151: What about authorization if device is password protected?

chemelli74: Am I wrong that OTA doesn't need auth ?

thecode: OTA needs auth same as other functions

aioshelly/common.py
```diff
@@ -137,3 +139,29 @@ def shelly_supported_firmware(result: dict[str, Any]) -> bool:
...
+    session: aiohttp.ClientSession,
+    host: str,
+    gen: int,
+    auth: aiohttp.BasicAuth | None = None,
+) -> bool:
+    """Trigger a firmware update via OTA http endpoint."""
+    if gen in BLOCK_GENERATIONS:
+        path = "ota"
+    else:
+        path = "rpc/Shelly.Update"
```
bdraco, line 154:
```suggestion
        path = "/ota"
    else:
        path = "/rpc/Shelly.Update"
```

aioshelly/common.py
```diff
@@ -137,3 +139,29 @@ def shelly_supported_firmware(result: dict[str, Any]) -> bool:
...
+    try:
+        await session.get(
+            URL.build(scheme="http", host=host, path=f"/{path}"),
```
bdraco, line 157:
```suggestion
            URL.build(scheme="http", host=host, path=path),
```

aioshelly/common.py
```diff
@@ -137,3 +139,29 @@ def shelly_supported_firmware(result: dict[str, Any]) -> bool:
...
+            auth=auth,
+            timeout=HTTP_CALL_TIMEOUT,
+        )
+    except aiohttp.ClientResponseError:
+        return False
+    except CONNECT_ERRORS:
+        return False
+
+    _LOGGER.debug("Update in progress for device %s", host)
+    return True
```
bdraco, line 167: It might be a bit nicer to throw a named exception so we can provide a better error in the UI if the update fails for some reason

aioshelly/common.py
bdraco: It would be nicer to avoid formatting again below

thecode:
I still think we should not go forward with this, the added complexity doesn't worth the benefit. If we complicated the code so much, **I would rather go back to were we was in core 2023.12 before we introduced the minimum firmware requirement (which was meant to reduce code complexity and not increase it.** 

Auth for Gen2 will not work as implemented here since Gen2 uses digest, if we have to duplicate this logic for HTTP we now increase the code that we have to maintain backward compatibility.

aioshelly/common.py
```diff
@@ -137,3 +139,29 @@ def shelly_supported_firmware(result: dict[str, Any]) -> bool:
...
+) -> bool:
+    """Trigger a firmware update via OTA http endpoint."""
+    if gen in BLOCK_GENERATIONS:
+        path = "ota"
+    else:
+        path = "rpc/Shelly.Update"
+    try:
+        await session.get(
+            URL.build(scheme="http", host=host, path=f"/{path}"),
+            auth=auth,
```
thecode, line 158: This will fail if used for Gen2

chemelli74: To avoid complexity of the code, we should move the Auth management out of the class so to reuse it here and there

thecode: The auth code is part of a class so you would have to do many changes for this and you end up with a code that was supposed to serve for wsrpc under common while it is nothing in common with Gen1.

I really don't understand were you are going with it. We enforce a minimum version which blocks us from initializing a device (although any minimum you set for Gen2 will not block you from initializing the device since there are no breaking API changes) than we move code to common so we can talk to the device without initializing it while we shouldn't have gone there in the first place.

chemelli74:
Superseded by #544

PR #487, opened 2024-01-20
Lower BLE 2.0 script max queue size to 32

bdraco:
80 proved to be to large to avoid running out of memory


chemelli74 (approved):

thecode (approved):
Updated prod, looks ok

bdraco:
thanks

PR #500, opened 2024-02-16
Introduce common library for all tools

chemelli74:


tools/common.py
```diff
@@ -0,0 +1,127 @@
...
+from aioshelly.const import BLOCK_GENERATIONS, MODEL_NAMES, RPC_GENERATIONS
+from aioshelly.exceptions import InvalidAuthError, ShellyError
+from aioshelly.rpc_device import RpcDevice, RpcUpdateType, WsServer
+
+coap_context = COAP()
+ws_context = WsServer()
+
+
+class CommonTools:
+    """Common method for building tools."""
```
bdraco, line 22: I don't think this needs to be a class as its doesn't ever set anything on `self`.  It looks like be a simple library of functions.

chemelli74: Done

bdraco (approved):

PR #501, opened 2024-02-16
Add retry to Block HTTP requests

thecode:
Since last week I have some failures with Gen1 devices, I still don't know if this is related to the transition to Python 3.12, a problem in my network or device problem. I suspect something happens faster and devices with ECO mode enabled fail to respond.

However even if it is a network or device problem, adding a single retry can improve the situation. We had many users complaining about failed actions, mainly to sleepy devices such as Shelly TRV & Shelly Motion.

I didn't increase the timeout, the retry still happen within our current timeout (10 seconds). For the last few days I had 33 retry occurrences and they all succeeded (meaning a single retry was enough, and without it 33 actions would fail). 
Retry will cause a short delay in action (up to two seconds) but is still acceptable from user experience and is better than a  failure.

### DIFF
--- a/aioshelly/block_device/device.py
+++ b/aioshelly/block_device/device.py
@@ -278,17 +278,18 @@ class BlockDevice:
         return event
 
     async def http_request(
-        self, method: str, path: str, params: Any | None = None
+        self, method: str, path: str, params: Any | None = None, retry: bool = True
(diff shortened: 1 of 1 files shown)


bieniu (approved):
Looks good 👍

PR #505, opened 2024-02-20
Allow setting of CoAP listening IP address

### LINKED ISSUE #149: aioshelly mcast ignores network interfaces settings
With the current implementation the multicast group membership will always be added to the interface that holds the default route. In my case this is wrong, since my system is multi-homed with default route pointing to eth0 and my Shelly's will communicate via eth1. As a result CoIoT notifications won't work and HA will fall back to polling.

See below the IGMP memberships on my system, with group BB0100E0 (which translates to 224.0.1.187) configured on eth0 when it should be on eth1:
```bash
bash-5.1# cat /proc/net/igmp
Idx	Device    : Count Querier	Group    Users Timer	Reporter
1	lo        :     1      V3
				010000E0     1 0:00000000		0
380	eth0      :     2      V3
				BB0100E0     1 0:00000000		0
				010000E0     1 0:00000000		0
378	eth1      :     3      V3
				FAFFFFEF     2 0:00000000		0
				FB0000E0     1 0:00000000		0
				010000E0     1 0:00000000		0
379	eth2      :     1      V3
				010000E0     1 0:00000000		0
```

The source code suggests the group membership should be configured on all interfaces (since IP_ADD_MEMBERSHIP is called with group address and IPADDR_ANY) so it might be a bug in Python, too, although I'm not aware of any at the moment.

My host is ArchLinux, same behaviour can be observed with current ghcr.io/home-assistant/home-assistant:stable (at the time of writing 2021.10.6)
```
# uname -a
Linux redblade 5.14.8-arch1-1 #1 SMP PREEMPT Sun, 26 Sep 2021 19:36:15 +0000 x86_64 GNU/Linux
# python --version 
Python 3.9.7
```

Adding the following code ("172.16.238.3" is eth1 interface address)
```python
sock.setsockopt(socket.IPPROTO_IP, socket.IP_ADD_MEMBERSHIP, socket.inet_aton("224.0.1.187") + socket.inet_aton("172.16.238.3"))
```
below line 66 (refers to code below) solves the issue for me, although this hardly qualifies as workaround. A better solution would be to iterate over interface addresses as configured by the user in Home Assistant settings -> Server control -> General -> Networks sections.

https://github.com/home-assistant-libs/aioshelly/blob/b1deaa2484c1192946588b0b3e8db263f8fde999/aioshelly/coap.py#L59-L68

chemelli74:
Home Assistant allows users to specify the internal network they want to use.
This makes integrations listen on that specific adapter instead of 0.0.0.0.

This PR allows to specify a IP address for CoAP listening, while not chancing current behaviour.

Fixes #149

### DIFF
--- a/aioshelly/block_device/coap.py
+++ b/aioshelly/block_device/coap.py
@@ -8,10 +8,11 @@ import socket
 import struct
 from collections.abc import Callable
 from enum import Enum, auto
+from ipaddress import IPv4Address
 from types import TracebackType
 from typing import cast
 
-from ..const import DEFAULT_COAP_PORT, DEFAULT_IP_ADDRESS
+from ..const import DEFAULT_COAP_PORT
 from ..json import JSONDecodeError, json_loads
 
 COAP_OPTION_DEVICE_ID = 3332
@@ -113,14 +114,23 @@ class CoapMessage:
         raise InvalidMessage("Option contained partial payload marker.")
 
 
-def socket_init(socket_ip: str, socket_port: int) -> socket.socket:
+def socket_init(
+    socket_ip: list[IPv4Address] | None = None,
+    socket_port: int = DEFAULT_COAP_PORT,
+) -> socket.socket:
     """Init UDP socket to send/receive data with Shelly devices."""
     sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM, socket.IPPROTO_UDP)
     sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
-    sock.bind((socket_ip, socket_port))
+    sock.bind(("", socket_port))
     _LOGGER.debug("Socket initialized on %s:%s", socket_ip, socket_port)
-    mreq = struct.pack("=4sl", socket.inet_aton("224.0.1.187"), socket.INADDR_ANY)
-    sock.setsockopt(socket.IPPROTO_IP, socket.IP_ADD_MEMBERSHIP, mreq)
+    if socket_ip:
+        for ip in socket_ip:
+            mreq = struct.pack("=4sl", socket.inet_aton("224.0.1.187"), ip)
+            sock.setsockopt(socket.IPPROTO_IP, socket.IP_ADD_MEMBERSHIP, mreq)
+    else:
+        # INADDR_ANY indicates that the OS will chose an interface to join the given multicast group.
+        mreq = struct.pack("=4sl", socket.inet_aton("224.0.1.187"), socket.INADDR_ANY)
+        sock.setsockopt(socket.IPPROTO_IP, socket.IP_ADD_MEMBERSHIP, mreq)
     sock.setblocking(False)
     return sock
 
@@ -137,7 +147,9 @@ class COAP(asyncio.DatagramProtocol):
         self.transport: asyncio.DatagramTransport | None = None
 
     async def initialize(
-        self, socket_ip: str = DEFAULT_IP_ADDRESS, socket_port: int = DEFAULT_COAP_PORT
+        self,
+        socket_ip: list[IPv4Address] | None = None,
+        socket_port: int = DEFAULT_COAP_PORT,
     ) -> None:
         """Initialize the COAP manager."""
         loop = asyncio.get_running_loop()
--- a/aioshelly/block_device/coap.py
+++ b/aioshelly/block_device/coap.py
@@ -122,18 +122,31 @@ def socket_init(
     sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM, socket.IPPROTO_UDP)
     sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
     sock.bind(("", socket_port))
-    if socket_ips:
-        for address in socket_ips:
-            ip = address.exploded
-            _LOGGER.debug("Socket initialized on %s:%s", ip, socket_port)
-            mreq = socket.inet_aton("224.0.1.187") + socket.inet_aton(ip)
-        sock.setsockopt(socket.IPPROTO_IP, socket.IP_ADD_MEMBERSHIP, mreq)
-    else:
+    sock.setblocking(False)
+    multicast_ip_bytes = socket.inet_aton("224.0.1.187")
+
+    if not socket_ips:
         _LOGGER.debug("Socket initialized on port %s (default interface)", socket_port)
         # INADDR_ANY indicates that the OS will chose an interface to join the given multicast group.
-        mreq = struct.pack("=4sl", socket.inet_aton("224.0.1.187"), socket.INADDR_ANY)
+        mreq = struct.pack("=4sl", multicast_ip_bytes, socket.INADDR_ANY)
         sock.setsockopt(socket.IPPROTO_IP, socket.IP_ADD_MEMBERSHIP, mreq)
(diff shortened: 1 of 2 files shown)


aioshelly/block_device/coap.py
```diff
@@ -111,14 +114,25 @@ def _read_extended_field_value(value: int, raw_data: bytes) -> tuple[int, bytes]
...
     sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
     sock.bind(("", socket_port))
-    _LOGGER.debug("Socket initialized on port %s", socket_port)
-    mreq = struct.pack("=4sl", socket.inet_aton("224.0.1.187"), socket.INADDR_ANY)
-    sock.setsockopt(socket.IPPROTO_IP, socket.IP_ADD_MEMBERSHIP, mreq)
+    if socket_ips:
+        mreq = socket.inet_aton("224.0.1.187")
+        for address in socket_ips:
+            _LOGGER.debug("Socket initialized on %s:%s", address, socket_port)
+            mreq += socket.inet_aton(address.exploded)
```
bdraco, line 129: This looks like its adding the ip addresses together?

bdraco: I think https://github.com/python-zeroconf/python-zeroconf/blob/4877829e6442de5426db152d11827b1ba85dbf59/src/zeroconf/_utils/net.py#L261 might be a good reference for how this should work

aioshelly/block_device/coap.py
```diff
@@ -111,14 +114,25 @@ def _read_extended_field_value(value: int, raw_data: bytes) -> tuple[int, bytes]
...
     sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
     sock.bind(("", socket_port))
-    _LOGGER.debug("Socket initialized on port %s", socket_port)
-    mreq = struct.pack("=4sl", socket.inet_aton("224.0.1.187"), socket.INADDR_ANY)
-    sock.setsockopt(socket.IPPROTO_IP, socket.IP_ADD_MEMBERSHIP, mreq)
+    if socket_ips:
+        for address in socket_ips:
+            ip = address.exploded
+            _LOGGER.debug("Socket initialized on %s:%s", ip, socket_port)
+            mreq = socket.inet_aton("224.0.1.187") + socket.inet_aton(ip)
```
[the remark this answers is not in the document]
chemelli74: Maybe, will update code in a while.
Strange is that current PR code works fine with 3 interfaces.

aioshelly/block_device/coap.py
```diff
@@ -111,14 +114,25 @@ def _read_extended_field_value(value: int, raw_data: bytes) -> tuple[int, bytes]
...
+) -> socket.socket:
     """Init UDP socket to send/receive data with Shelly devices."""
     sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM, socket.IPPROTO_UDP)
     sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
     sock.bind(("", socket_port))
-    _LOGGER.debug("Socket initialized on port %s", socket_port)
-    mreq = struct.pack("=4sl", socket.inet_aton("224.0.1.187"), socket.INADDR_ANY)
-    sock.setsockopt(socket.IPPROTO_IP, socket.IP_ADD_MEMBERSHIP, mreq)
+    if socket_ips:
+        for address in socket_ips:
```
bdraco, line 126: I guess we need to catch OSError here and log like we do in zeroconf and only raise if all of them fail

https://github.com/python-zeroconf/python-zeroconf/blob/4877829e6442de5426db152d11827b1ba85dbf59/src/zeroconf/_utils/net.py#L289

aioshelly/block_device/coap.py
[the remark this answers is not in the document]
bdraco: I think you can write the function as 
```python
def socket_init(
    socket_ips: list[IPv4Address] | None = None,
    socket_port: int = DEFAULT_COAP_PORT,
) -> socket.socket:
    """Init UDP socket to send/receive data with Shelly devices."""
    sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM, socket.IPPROTO_UDP)
    sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
    sock.bind(("", socket_port))
    sock.setblocking(False)
    multicast_ip_bytes = socket.inet_aton("224.0.1.187")

    if not socket_ips:
        _LOGGER.debug("Socket initialized on port %s (default interface)", socket_port)
        # INADDR_ANY indicates that the OS will chose an interface to join the given multicast group.
        mreq = struct.pack("=4sl", multicast_ip_bytes, socket.INADDR_ANY)
        sock.setsockopt(socket.IPPROTO_IP, socket.IP_ADD_MEMBERSHIP, mreq)
        return sock

    had_successful_joins = False
    last_exception: OSError | None = None
    for address in socket_ips:
        _LOGGER.debug("Socket initialized on %s:%s", address, socket_port)
        mreq = multicast_ip_bytes + address.packed
        try:
            sock.setsockopt(socket.IPPROTO_IP, socket.IP_ADD_MEMBERSHIP, mreq)
            had_successful_joins = True
        except OSError as ex:
            _LOGGER.warning("Failed to join multicast group on %s", address)
            last_exception = ex

    if not had_successful_joins and last_exception is not None:
        raise last_exception

    return sock
```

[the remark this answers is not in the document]
chemelli74: Tested and working. Committed. Thx!

bdraco (approved):

aioshelly/block_device/coap.py
```diff
@@ -134,10 +161,14 @@ def __init__(self, message_received: Callable | None = None) -> None:
         self.subscriptions: dict[str, Callable] = {}
         self.transport: asyncio.DatagramTransport | None = None
 
-    async def initialize(self, socket_port: int = 5683) -> None:
+    async def initialize(
+        self,
+        socket_ips: list[IPv4Address] | None = None,
+        socket_port: int = DEFAULT_COAP_PORT,
+    ) -> None:
```
thecode, line 168: With this change we are creating a breaking API change for `aioshelly` users and if we want to use this for older versions we won't be able too. if we add the `socket_ips` as a last parameter there are no breaking changes
```suggestion
    async def initialize(
        self,
        socket_port: int = DEFAULT_COAP_PORT,
        socket_ips: list[IPv4Address] | None = None,
    ) -> None:
```

thecode:
It would be nice if this PR will also adjust the https://github.com/home-assistant-libs/aioshelly/blob/main/tools/example.py (and maybe other tools which are affected) to support setting the listening IP so this change can be used as a standalone and for testing/fixture dumping when we want to use a different listening address as added in this PR

aioshelly/block_device/coap.py
chemelli74: Good point Shay, let me switch them!

chemelli74:
> It would be nice if this PR will also adjust the https://github.com/home-assistant-libs/aioshelly/blob/main/tools/example.py (and maybe other tools which are affected) to support setting the listening IP so this change can be used as a standalone and for testing/fixture dumping when we want to use a different listening address as added in this PR

I will take care in a new PR of the tools.

thecode (approved):
Thanks!

PR #509, opened 2024-02-27
Simplify logging during debugging

chemelli74:
Remove background noise from logging when debug option is enabled.

**Background**:

When you are debugging gen2, having all debug traffic from gen1 is just filling quickly the window making easy to miss what you need.
And the other way around with gen2.

**Solution**:

So if you specify a gen as arg, and only on that case, you just lower the logging for others gen in order to focus on what you are looking for

### DIFF
--- a/tools/example.py
+++ b/tools/example.py
@@ -199,7 +199,11 @@ async def main() -> None:
         gen = 3
 
     if args.debug:
-        logging.basicConfig(level="DEBUG", force=True)
+        logging.basicConfig(level=logging.DEBUG)
+        if args.gen1:
+            logging.getLogger("aioshelly.rpc_device").setLevel(logging.INFO)
+        else:
+            logging.getLogger("aioshelly.block_device").setLevel(logging.INFO)
 
     def handle_sigint(_exit_code: int, _frame: FrameType) -> None:
         """Handle Keyboard signal interrupt (ctrl-c)."""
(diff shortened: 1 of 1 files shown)


tools/example.py
```diff
@@ -199,7 +199,11 @@ async def main() -> None:
         gen = 3
 
     if args.debug:
-        logging.basicConfig(level="DEBUG", force=True)
+        logging.basicConfig(level=logging.DEBUG)
+        if args.gen1:
+            logging.getLogger("aioshelly.rpc_device").setLevel(logging.INFO)
+        else:
```
thecode, line 205: The current implementation doesn't match the PR description, if you don't specify any gen you disable debug logging for Gen1, should check that no Gen is in params:
```suggestion
        elif args.gen2 or args.gen3:
```

tools/example.py
```diff
@@ -199,7 +199,11 @@ async def main() -> None:
         gen = 3
 
     if args.debug:
-        logging.basicConfig(level="DEBUG", force=True)
+        logging.basicConfig(level=logging.DEBUG)
+        if args.gen1:
```
thecode, line 203:
```suggestion
        # if gen is in args reduce logging for other gens
        if args.gen1:
```

thecode (approved):

chemelli74:
Thx!

PR #512, opened 2024-02-27
Fix backward params compatibility

chemelli74:
Change parameters order to allow backward compatibility


chemelli74:
Confirmed to be working with current code

bdraco (approved):

PR #513, opened 2024-02-27
Save strong reference to create_task

bdraco:
> Important Save a reference to the result of this function, to avoid a task disappearing mid-execution. The event loop only keeps weak references to tasks. A task that isn’t referenced elsewhere may get garbage collected at any time, even before it’s done. For reliable “fire-and-forget” background tasks, gather them in a collection:


bieniu (approved):
Tested 👍

bdraco:
thanks

PR #515, opened 2024-03-05
Update example to listen to specified ip address for CoAP packets

chemelli74:


tools/example.py
```diff
@@ -186,6 +187,12 @@ def get_arguments() -> tuple[argparse.ArgumentParser, argparse.Namespace]:
         type=str,
         help="Update outbound WebSocket (Gen2/3) and exit",
     )
+    parser.add_argument(
+        "--listen_ip_address",
+        "-lip",
+        type=str,
+        help="Listen ip address for incoming CoAP packets",
```
thecode, line 194:
```suggestion
        type=str,
        nargs='*',
        default=None,        
        help="Listen ip address for incoming CoAP packets",
```

With `nargs='*'` output will be a list and default can be set to `None` if no input
https://docs.python.org/3/library/argparse.html#nargs

Example output of the above with different input
`example.py --ip 1.2.2.3`  --> `args.listen_ip_address=None`
`example.py --ip 1.2.2.3 -lip 1.2.3.4` -->  `args.listen_ip_address=['1.2.3.4']`
`example.py --ip 1.2.2.3 -lip 1.2.3.4 5.6.7.8` --> `args.listen_ip_address=['1.2.3.4', '5.6.7.8']`

tools/example.py
```diff
@@ -196,7 +203,10 @@ async def main() -> None:
     """Run main."""
     parser, args = get_arguments()
 
-    await coap_context.initialize(args.coap_port)
+    listen_ip_address = (
+        [IPv4Address(args.listen_ip_address)] if args.listen_ip_address else None
+    )
+    await coap_context.initialize(args.coap_port, listen_ip_address)
```
thecode, line 209:
```suggestion
    await coap_context.initialize(args.coap_port, args.listen_ip_address)
```

tools/example.py
chemelli74: Nice, thx !

thecode (approved):

PR #523, opened 2024-03-17
Add model name for Plus Plug S V2

bieniu:
obraz

https://kb.shelly.cloud/knowledge-base/shelly-plus-plug-s-v2


thecode (approved):

aioshelly/const.py
```diff
@@ -168,6 +169,7 @@
     MODEL_PLUS_I4DC: "Shelly Plus I4DC",
     MODEL_PLUS_PLUG_IT: "Shelly Plus Plug IT",
     MODEL_PLUS_PLUG_S: "Shelly Plus Plug S",
+    MODEL_PLUS_PLUG_S_V2: "Shelly Plus Plug S",
```
chemelli74, line 172:
```suggestion
    MODEL_PLUS_PLUG_S_V2: "Shelly Plus Plug S V2",
```

chemelli74 (changes requested):
I think we should specify that it's a V2 model

aioshelly/const.py
bieniu: In the past, we did not add the V2 suffix for new hardware versions. I think we should be consistent.

https://github.com/home-assistant-libs/aioshelly/blob/e8058a54a38276050b6bd297954e041ea35082e3/aioshelly/const.py#L189-L190

chemelli74: In the past there was no different name from Shelly for different revision; seems this time they changed approach as there are small differences

bieniu: I'll ask the Shelly team if this device will actually be sold as V2, if not I think we should not introduce the suffix as it will be confusing for users.

chemelli74 (approved):

bieniu:
Thanks @thecode @chemelli74

PR #526, opened 2024-03-20
Move call to initialize to user code

thecode:
Remove background `_async_init` task and `initialize` parameter:

- Move call to `initialize` for sleeping devices to user code, `_async_init` is no longer runs inside `aioshelly` which allows the user to call `initialize` when device is online and catch the errors.
- Add new status update type: `BlockUpdateType.ONLINE` & `RpcUpdateType.ONLINE` which inform the user that the device is online and he can call the `initialize`.
- Remove `initialize` parameter from device create. User is responsible to call `initialize`. This helps that we don't need to catch errors in two different places and can catch errors only on `initialize` method.
- Update examples

In the future we can use mDNS advertising from the device to call `initialize` which will help setting up sleeping devices which doesn't have `CoAP` unicast configured or `Outbound Websocket` so we can set it up for them.

### DIFF
--- a/aioshelly/block_device/device.py
+++ b/aioshelly/block_device/device.py
@@ -115,11 +115,6 @@ class BlockDevice:
         """Device ip address."""
         return self.options.ip_address
 
-    @property
-    def port(self) -> int:
-        """Device port."""
-        return DEFAULT_HTTP_PORT
-
     async def initialize(self) -> None:
         """Device initialization."""
         if self._initializing:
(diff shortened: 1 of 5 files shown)


chemelli74:
What about splitting the port change to a separate PR ?

thecode:
> What about splitting the port change to a separate PR ?

I removed the port change from this PR

chemelli74 (approved):
LGTM, thx @thecode

bieniu (approved):
Looks good 👍
Tested with several devices and I don't see any problems.

thecode:
Thanks @chemelli74 @bieniu

PR #527, opened 2024-03-21
Drop python 3.10

chemelli74:


thecode (approved):
I checked for all the places in the repo with `3.10` and looks like all are updated

PR #532, opened 2024-03-22
Add pre-commit autoupdate GH action

bieniu:
Currently, we very rarely update the version of pre-commit hooks. For example, the `ruff` and `mypy` hooks are currently not up-to-date. I suggest adding GH action, which will run `pre-commit autoupdate` once a week and open PR if updates are available.

### DIFF
--- a/.github/workflows/pre-commit-update.yml
+++ b/.github/workflows/pre-commit-update.yml
@@ -1,0 +1,34 @@
+name: Pre-commit auto-update
+on:
+  schedule:
+    # Run on mondays at midnight
+    - cron: "0 0 * * 1"
+
+jobs:
+  auto-update:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4.1.2
+
+      - name: Set up Python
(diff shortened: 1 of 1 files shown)


.github/workflows/pre-commit-update.yml
```diff
@@ -0,0 +1,34 @@
...
+    - cron: "0 0 * * 1"
+
+jobs:
+  auto-update:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4.1.2
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
```
thecode, line 14:
```suggestion
        uses: actions/setup-python@v5.0.0
```
It would be good to pin `setup-python` we do the same in other workflows:
https://github.com/home-assistant-libs/aioshelly/blob/497b7a44cc954198bfba4e98a62c163bf5ffc841/.github/workflows/pythonpublish.yml#L17

thecode (approved):

chemelli74 (approved):
Thx!

PR #533, opened 2024-03-22
Update RPC shutdown description

thecode:
The RPC shutdown process is not reversible since there is no method to reconnect the websocket.
Remove the comment to avoid incorrectly trying to reverse shutdown.


bieniu (approved):

chemelli74 (approved):

thecode:
Thanks!

PR #534, opened 2024-03-22
Add comment to explain the usage of the aioshelly BLE script

thecode:
Shelly is getting reports from users not able to delete a specific script and suggested to add a comment on top of the script body stating clearly that it is installed by HA


bieniu (approved):

chemelli74 (approved):

thecode:
Thanks!

PR #535, opened 2024-03-24
Prevent KeyError on device shutdown

thecode:
Prevent raising `KeyError` if shutdown is in incorrect order and log it as an error:
Will log:
```
host 192.168.1.135: error during shutdown: KeyError('55960A')
```
Since we use a shared CoAP/ WsRPC context devices can cause error if not being shutdown at the correct order.

The following sequence (modified example.py to create two Gen2 devices) will raise a `KeyError`:

```python
            device1 = await create_device(aiohttp_session, options, init, gen)
            device2 = await create_device(aiohttp_session, options, init, gen)
            await device2.shutdown()
            await device1.shutdown()
```
```
    self.device.shutdown()
  File "/usr/local/lib/python3.12/site-packages/aioshelly/block_device/device.py", line 186, in shutdown
    self._unsub_coap()
  File "/usr/local/lib/python3.12/site-packages/aioshelly/block_device/coap.py", line 233, in <lambda>
    return lambda: self.subscriptions.pop(ip_or_device_id)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
KeyError: '55960A'
```
While devices should be shutdown in the correct order even if they don't it would be good to allow the devices to shutdown without raising an error since the device being shutdown anyhow.

### DIFF
--- a/aioshelly/block_device/coap.py
+++ b/aioshelly/block_device/coap.py
@@ -230,7 +230,7 @@ class COAP(asyncio.DatagramProtocol):
         """Subscribe to received updates."""
         _LOGGER.debug("Adding device %s to CoAP message subscriptions", ip_or_device_id)
         self.subscriptions[ip_or_device_id] = message_received
-        return lambda: self.subscriptions.pop(ip_or_device_id, None)
+        return lambda: self.subscriptions.pop(ip_or_device_id)
 
     async def __aenter__(self) -> Self:
         """Entering async context manager."""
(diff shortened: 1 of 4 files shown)


aioshelly/block_device/coap.py
```diff
@@ -230,7 +230,7 @@ def subscribe_updates(
         """Subscribe to received updates."""
         _LOGGER.debug("Adding device %s to CoAP message subscriptions", ip_or_device_id)
         self.subscriptions[ip_or_device_id] = message_received
-        return lambda: self.subscriptions.pop(ip_or_device_id)
+        return lambda: self.subscriptions.pop(ip_or_device_id, None)
```
bdraco, line 233: I think we log and error here so we don't hide the problem if we recreate it in the future (races are hard)

bdraco (approved):

thecode:
Thanks!

PR #544, opened 2024-04-09
Allow initialization for device with unsupported firmware

thecode:
Allow initialization of a device with unsupported firmware so that we can use the device to trigger an update.
Added property to indicate if firmware is supported (`firmware_supported`)

Updated the `example.py` to indicate if firmware is not supported:

<ins>Gen1:</ins>
```
** shelly4pro-801234 - Shelly 4Pro  @ 192.168.1.49:80**

Device firmware not supported: 20190821-095050/v1.5.2@4148d2b7
```

<ins>Gen2:</ins>
```
** shellyplus1-a8032abe1234 - Shelly Plus 1  @ 192.168.1.70:80**

Device firmware not supported: 20210714-140932
```

(this Shelly pre-production firmware doesn't have a version only a date)

### DIFF
--- a/aioshelly/block_device/device.py
+++ b/aioshelly/block_device/device.py
@@ -147,7 +147,7 @@ class BlockDevice:
                 # Older devices has incompatible CoAP protocol (v1)
                 # Skip CoAP to avoid parsing errors
                 if self.firmware_supported:
-                    event_d: asyncio.Event = await self._coap_request("d")
+                    event_d = await self._coap_request("d")
                     # We need to wait for D to come in before we request S
                     # Or else we might miss the answer to D
                     await event_d.wait()
--- a/aioshelly/block_device/device.py
+++ b/aioshelly/block_device/device.py
@@ -17,13 +17,17 @@ from ..const import (
     CONNECT_ERRORS,
     DEFAULT_HTTP_PORT,
     DEVICE_IO_TIMEOUT,
+    FIRMWARE_PATTERN,
+    GEN1_LIGHT_TRANSITION_MIN_FIRMWARE_DATE,
+    GEN1_MIN_FIRMWARE_DATE,
+    GEN1_MODELS_SUPPORTING_LIGHT_TRANSITION,
+    GEN1_MODELS_UNSUPPORTED,
     HTTP_CALL_TIMEOUT,
     MODEL_RGBW2,
 )
 from ..exceptions import (
     CustomPortNotSupported,
     DeviceConnectionError,
-    FirmwareUnsupported,
     InvalidAuthError,
     MacAddressMismatchError,
     NotInitialized,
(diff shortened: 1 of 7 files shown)


aioshelly/block_device/device.py
```diff
@@ -77,7 +81,7 @@ def __init__(
         self.aiohttp_session: ClientSession = aiohttp_session
         self.options: ConnectionOptions = options
         self.coap_d: dict[str, Any] | None = None
-        self.blocks: list | None = None
+        self.blocks: list = []
```
bdraco, line 84: Would be nice to type the list

aioshelly/block_device/device.py
```diff
@@ -77,7 +81,7 @@ def __init__(
...
         self.coap_s: dict[str, Any] | None = None
         self._settings: dict[str, Any] | None = None
         self._shelly: dict[str, Any] | None = None
```
bdraco, line 87: I would have expected 

`self._coap_response_events: dict[str, asyncio.Event] = {}`

aioshelly/block_device/device.py
```diff
@@ -140,14 +144,17 @@ async def initialize(self) -> None:
...
                 await self.update_status()
 
-                event_d: asyncio.Event = await self._coap_request("d")
-                # We need to wait for D to come in before we request S
-                # Or else we might miss the answer to D
-                await event_d.wait()
+                # Older devices has incompatible CoAP protocol (v1)
+                # Skip CoAP to avoid parsing errors
+                if self.firmware_supported:
+                    event_d: asyncio.Event = await self._coap_request("d")
```
bdraco, line 150: This typing looks redundant
```suggestion
                    event_d = await self._coap_request("d")
```

aioshelly/rpc_device/device.py
```diff
@@ -412,3 +419,17 @@ def connected(self) -> bool:
     def last_error(self) -> ShellyError | None:
         """Return the last error during async device init."""
         return self._last_error
+
+    @property
+    def firmware_supported(self) -> bool:
+        """Return True if device firmware version is supported."""
+        fw_ver = GEN2_MIN_FIRMWARE_DATE if self.gen == GEN2 else GEN3_MIN_FIRMWARE_DATE
+
+        match = FIRMWARE_PATTERN.search(self.firmware_version)
```
bdraco, line 428: If we are going to call this often, it might make sense to make the firmware version parser a lru_cache decorated function.

If we aren't going to call it a lot it doesn't matter

bdraco (approved):

aioshelly/rpc_device/device.py
thecode: This is called once after we initialize a device

bdraco (approved):

thecode:
Thanks!

bieniu (approved):
Looks good, thanks @thecode

thecode:
Thanks @bieniu

PR #546, opened 2024-04-11
Code clean-ups

thecode:


aioshelly/rpc_device/wsrpc.py
```diff
@@ -50,21 +50,20 @@
 
 def _receive_json_or_raise(msg: WSMessage) -> dict[str, Any]:
     """Receive json or raise."""
+    if msg.type is WSMsgType.TEXT:
```
thecode, line 53: Most of the times we are going to get `WSMsgType.TEXT` message so it would be better to check this first

chemelli74 (approved):
LGTM

bieniu (approved):

thecode:
Thanks

PR #563, opened 2024-05-10
Add debug logging

thecode:
To help debug https://github.com/home-assistant/core/issues/116975 better

### DIFF
--- a/aioshelly/block_device/device.py
+++ b/aioshelly/block_device/device.py
@@ -112,7 +112,11 @@ class BlockDevice:
             await coap_context.request(options.ip_address, "s")
         except OSError as err:
             _LOGGER.debug("host %s: error: %r", options.ip_address, err)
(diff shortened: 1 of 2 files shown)


chemelli74:
What about adding also the MAC address to the log message ?
Just in case there is also DHCP kicking in between one log session and the next one

thecode:
> What about adding also the MAC address to the log message ? Just in case there is also DHCP kicking in between one log session and the next one

Added to device create:
```
DEBUG:aioshelly.block_device.device:host 192.168.1.135: block device create, MAC: CAAB12345ABC
DEBUG:aioshelly.rpc_device.device:host 192.168.1.35:80: RPC device create, MAC: 12345AB50124

```

bieniu (approved):

chemelli74 (approved):
Thx!

thecode:
Thanks!

PR #568, opened 2024-05-17
Align device shutdown

thecode:
Align device shutdown method so both block device and RPC device methods will be async, this makes it easier that devices has common interface (`create`, `initialize`, `subscribe_updates` & `shutdown`)

# Breaking change
For block device the method was changed to `async`, it should be awaited now.

## Before
```python
device.shutdown()
```

## After
```python
await device.shutdown()
```

Example code updated to call shutdown upon `ctrl-c` so we can test device shutdown.


bdraco (approved):

thecode:
Thanks!

PR #575, opened 2024-06-07
Fetch new status on device re-init

thecode:
Fix https://github.com/home-assistant/core/issues/118898

When a device status is updated while it is disconnected and we reconnect to the device we the status is not updated since we already have a status (this logic is needed for sleeping devices).

This makes sure that if we re-init a device we also fetch new status from it.

### DIFF
--- a/aioshelly/block_device/device.py
+++ b/aioshelly/block_device/device.py
@@ -136,7 +136,13 @@ class BlockDevice:
             raise CustomPortNotSupported
 
         self._initializing = True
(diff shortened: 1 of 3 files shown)


bdraco (approved):
LGTM

bdraco:
I disconnected a v1 and a v2 device from wifi in unifi network and verified they came back ok

thecode:
Thanks @bdraco

PR #583, opened 2024-06-18
Add support for virtual components

bieniu:
Virtual components docs: https://shelly-api-docs.shelly.cloud/gen2/DynamicComponents/Virtual/

### DIFF
--- a/aioshelly/rpc_device/device.py
+++ b/aioshelly/rpc_device/device.py
@@ -470,8 +470,3 @@ class RpcDevice:
             for component in result["components"]
             if any(supported in component["key"] for supported in VIRTUAL_COMPONENTS)
         ]
-
-    @property
-    def dynamic_components(self) -> list[dict[str, Any]]:
-        """Return a list of dynamic components."""
-        return self._dynamic_components
--- a/aioshelly/rpc_device/device.py
+++ b/aioshelly/rpc_device/device.py
@@ -19,6 +19,7 @@ from ..const import (
     GEN2_MIN_FIRMWARE_DATE,
     GEN3_MIN_FIRMWARE_DATE,
     NOTIFY_WS_CLOSED,
+    VIRTUAL_COMPONENTS,
 )
 from ..exceptions import (
     DeviceConnectionError,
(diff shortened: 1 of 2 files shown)


aioshelly/rpc_device/device.py
```diff
@@ -252,10 +255,19 @@ async def trigger_reboot(self, delay_ms: int = 1000) -> None:
     async def update_status(self) -> None:
         """Get device status from 'Shelly.GetStatus'."""
         self._status = await self.call_rpc("Shelly.GetStatus")
+        self._status.update(
+            {
+                item["key"]: {"value": item["status"].get("value")}
```
bieniu, line 260: `button` virtual component has no `value` in the `status` so we need use `get()` here.

thecode: I think it would be cleaner to move this to line 193 (after we fetch status) and make `get_dynamic_components` update the status and the config right after fetching the data

aioshelly/rpc_device/device.py
```diff
@@ -249,10 +252,19 @@ async def trigger_reboot(self, delay_ms: int = 1000) -> None:
...
+                for item in self._dynamic_components
+            }
+        )
 
     async def update_config(self) -> None:
         """Get device config from 'Shelly.GetConfig'."""
         self._config = await self.call_rpc("Shelly.GetConfig")
+        self._config.update(
+            {item["key"]: item["config"] for item in self._dynamic_components}
+        )
```
thecode, line 267: Would be cleaner if `get_dynamic_components` will do this so we have all in a single place when we get the new data

aioshelly/rpc_device/device.py
thecode: Just notice we fetch status at
https://github.com/home-assistant/core/blob/ac8ea8356c8483c2123ba9c12c3cd1b7652093ab/homeassistant/components/shelly/coordinator.py#L741

So if we do the change I suggested we will need to call `get_dynamic_components` there so maybe also

thecode (approved):
Looks good, Thanks!

chemelli74 (approved):
LGTM !

PR #611, opened 2024-08-08
Revert "Significantly reduce heartbeat overhead (#301)"

### LINKED ISSUE #610: Revert #301 once aiohttp timer handle churn is reduced
Once https://github.com/aio-libs/aiohttp/pull/8608 merges we can remove the custom ping/pong implementation as the issue will be solved upstream

```[tasklist]
### Tasks
- [x] Wait for aiohttp 3.10.2 release
- [x] Wait for aiohttp to be bumped to 3.10.2
- [x] Rework a revert PR for #301
```

bdraco:
This reverts commit e3020ff49fcf10438ce923ea53bc82a710684070.

This change is no longer needed after https://github.com/aio-libs/aiohttp/pull/8608

closes #610

### DIFF
--- a/aioshelly/rpc_device/wsrpc.py
+++ b/aioshelly/rpc_device/wsrpc.py
@@ -43,8 +43,6 @@ from ..json import json_dumps, json_loads
 
 _LOGGER = logging.getLogger(__name__)
 
-WS_HEATBEAT_HALF_INTERVAL = WS_HEARTBEAT / 2
-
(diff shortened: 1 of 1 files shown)


thecode (approved):
Looks good, I think it worth waiting for https://github.com/home-assistant-libs/aioshelly/pull/613 since we might release a bugfix before that.

bieniu (approved):
Tested with real device, works as intended.

PR #612, opened 2024-08-08
Use aiohttp.ClientTimeout for passing timeouts

bdraco:
Calling timeout with float is deprecated, and the typing now reflects the deprecation. use aiohttp.ClientTimeout instead


thecode (approved):
Thanks!

PR #613, opened 2024-08-09
Don't use `GetComponents` if firmware doesn't support it

bieniu:
Related to https://github.com/home-assistant/core/issues/123440

The integration uses `component_added` and `component_removed` events so firmware 1.2.0 is the minimum version to ensure full compatibility with virtual components.

obraz

Tested with firmware 1.1.0 and 1.2.0.


bieniu:
It's not ready yet, I'll test the change with the old firmware in the evening.

thecode (approved):

PR #618, opened 2024-08-15
Add a lock for RPC device initialization

bdraco:
In https://github.com/home-assistant/core/issues/123399 we see there is a race after the device drops the
connection

### DIFF
--- a/aioshelly/rpc_device/device.py
+++ b/aioshelly/rpc_device/device.py
@@ -82,9 +82,7 @@ class RpcDevice:
         self._event: dict[str, Any] | None = None
         self._config: dict[str, Any] | None = None
         self._dynamic_components: list[dict[str, Any]] = []
-        self._wsrpc = WsRPC(
-            options.ip_address, self._on_notification, port=options.port
-        )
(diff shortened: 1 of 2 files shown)


aioshelly/rpc_device/device.py
```diff
@@ -206,11 +217,12 @@ async def initialize(self) -> None:
...
-        finally:
-            self._initializing = False
-
-        if self._update_listener and self.initialized:
-            self._update_listener(self, RpcUpdateType.INITIALIZED)
+        else:
+            _LOGGER.debug("host %s:%s: RPC device init finished", ip, port)
+            # Set notification callback once we are finished
+            # the init process
+            self._wsrpc.set_notification_callback(self._on_notification)
```
bdraco, line 224: We don't accept notifications until after https://github.com/home-assistant-libs/aioshelly/pull/618/files#diff-b3a38f3aae5db155a39278d8fad4bd7b03c9255b5fa3dff2ae6d072fccf83b7eR207 is done.

I think thats ok, but someone should double check that makes sense

bieniu (approved):
Tested with gen2 device, no issues.

bdraco:
thanks

PR #619, opened 2024-08-15
Optimize RPC device dict merge

bdraco:
I noticed the dict merge makes two copies of the destination dict, one inside `mergedicts` and one outside.  It doesn't look like we need to make a make a copy in the first place since we never compare the old and new data anywhere.

This is a ~92% speed up

dict_merge


bieniu (approved):
Looks good

bdraco:
Thanks

PR #622, opened 2024-08-17
Fix battery powered devices not getting ONLINE messages

bdraco:
#618 prevented battery powered devices from getting the ONLINE message because it did not connect the
the notification callback until after the client websocket init process had completed.

To fix this
- Revert to connecting the notification callback right away
- Only convert notify messages to ONLINE when they are recieved via the server. The client side should never convert notify to ONLINE as it will trigger Home Assistant to try to reconnect again when the connection is already established. This would happen most commonly with BLE notify because the script could still be running during a reconnect which would generate these notify messages before the device had finished initializing.

### DIFF
--- a/aioshelly/rpc_device/device.py
+++ b/aioshelly/rpc_device/device.py
@@ -6,6 +6,7 @@ import asyncio
 import logging
 from collections.abc import Callable
 from enum import Enum, auto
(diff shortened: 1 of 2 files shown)


thecode (approved):
Tested with HT Gen2

PR #623, opened 2024-08-17
Save a strong reference to the cleanup task in WsServer

bdraco:
followup fix to #405


thecode (approved):
Tested with Shelly Plus 1PM & Shelly HT Gen2.

bdraco:
Thanks

PR #632, opened 2024-08-20
Check if firmware is supported by device model

bieniu:


aioshelly/const.py
```diff
@@ -118,109 +121,181 @@
...
-    MODEL_SENSE: "Shelly Sense",
-    MODEL_SMOKE: "Shelly Smoke",
-    MODEL_SMOKE_2: "Shelly Smoke 2",
-    MODEL_SPOT: "Shelly Spot",
-    MODEL_SPOT_2: "Shelly Spot 2",
-    MODEL_UNI: "Shelly UNI",
-    MODEL_VALVE: "Shelly Valve",
-    MODEL_VINTAGE: "Shelly Vintage",
-    MODEL_VINTAGE_V2: "Shelly Vintage",
+    MODEL_1: {ATTR_MODEL_NAME: "Shelly 1", ATTR_MIN_FW_DATE: 20201124},
```
bdraco, line 126: I'd prefer a dataclass here instead of a dict

tests/test_const.py
```diff
@@ -0,0 +1,20 @@
+"""Tests for common module."""
```
bdraco, line 1:
```suggestion
"""Tests for const module."""
```

thecode (approved):
Thanks @bieniu @bdraco

PR #633, opened 2024-08-25
Reduce overhead to make RPC calls

bdraco:
We can avoid the asyncio.timeout context since we already are awaiting a future and schedule a timeout on the future that is being awaited instead. This saves a bit of latency which helps with being able to connect in time to sleepy devices for the few seconds they are awake.

### DIFF
--- a/aioshelly/rpc_device/wsrpc.py
+++ b/aioshelly/rpc_device/wsrpc.py
@@ -298,7 +298,7 @@ class WsRPC(WsBase):
                 return
 
             call = self._calls.pop(frame_id)
-            if not call.resolve.cancelled():
+            if not call.resolve.done():
                 call.resolve.set_result(frame)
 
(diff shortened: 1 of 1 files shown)


aioshelly/rpc_device/wsrpc.py
```diff
@@ -390,16 +396,17 @@ async def _rpc_call(
...
         self._calls[call.call_id] = call
+        timeout_handle = self._loop.call_at(
+            self._loop.time() + timeout, _handle_timeout, future
+        )
 
         try:
-            async with asyncio.timeout(timeout):
-                await self._send_json(call.request_frame)
-                resp = await future
+            await self._send_json(call.request_frame)
```
bdraco, line 404: The problem with this approach is that the json send could block forever if we hit the high water mark

PR #638, opened 2024-08-26
Add support for parallel rpc calls

bdraco:
Currently all calls during the websocket reconnect process happened in sequence:

 send -> wait -> read -> send -> wait -> read -> send -> wait -> read ...

This pattern results in the reconnect process taking a bit longer.

We can now send multiple calls and await for their responses:

 send all -> wait all -> read all

This is not implemented as parallel tasks as there is a chance we need to retry requests if auth needs to be recalculated

### DIFF
--- a/aioshelly/rpc_device/wsrpc.py
+++ b/aioshelly/rpc_device/wsrpc.py
@@ -132,7 +132,7 @@ class RPCCall:
         "dst",
         "resolve",
         "result",
-        "has_result",
+        "success",
     )
 
     def __init__(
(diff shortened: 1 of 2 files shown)


aioshelly/rpc_device/wsrpc.py
```diff
@@ -357,52 +374,129 @@ async def call(
...
-            with contextlib.suppress(asyncio.CancelledError):
-                future.cancel()
-                await future
+            for call in sent_calls:
+                with contextlib.suppress(asyncio.CancelledError):
+                    call.resolve.cancel()
+                    await call.resolve
+                # Ensure the call is removed from the calls dict
+                # on failure
+                self._calls.pop(call.call_id, None)
```
bdraco, line 485: This is technically a bug fix but its probably not a real problem since if we have timeouts it generally means the connection is going to get torn down anyways and the dict will be cleared

bieniu (approved):
Looks good, tested with WallDisplay, Pro 1 and Plus HT with and without auth.

thecode (approved):
Tested with Shelly Pro 2PM and Shelly Plus HT
In addition it is running on my prod with various types of devices for few hours

PR #639, opened 2024-08-26
Switch fetching device info to go over the websocket for RPC devices

bdraco:
Avoid the REST call since we have to create another connection to the device when we do the rest call which will stay open for a bit. By doing everything of the websocket we put less pressure on the device, and setup is a bit faster

### DIFF
--- a/aioshelly/rpc_device/device.py
+++ b/aioshelly/rpc_device/device.py
@@ -285,18 +285,18 @@ class RpcDevice:
 
     async def _init_calls(self) -> None:
         """Make calls needed to initialize the device."""
-        calls: list[tuple[str, dict[str, Any] | None]] = [("Shelly.GetConfig", None)]
-        if fetch_status := self._status is None:
-            calls.append(("Shelly.GetStatus", None))
+        calls: list[tuple[str, dict[str, Any] | None]] = []
         if fetch_dynamic := self._supports_dynamic_components():
             calls.append(("Shelly.GetComponents", {"dynamic_only": True}))
+        if fetch_status := self._status is None:
+            calls.append(("Shelly.GetStatus", None))
+        calls.append(("Shelly.GetConfig", None))
         results = await self.call_rpc_multiple(calls)
-        self._config = results[0]
-        if fetch_status:
-            self._status = results[1]
         if fetch_dynamic:
-            dynamic_idx = 2 if fetch_status else 1
-            self._parse_dynamic_components(results[dynamic_idx])
+            self._parse_dynamic_components(results.pop())
+        if fetch_status:
+            self._status = results.pop()
+        self._config = results[0]
 
     async def script_list(self) -> list[ShellyScript]:
         """Get a list of scripts from 'Script.List'."""
--- a/aioshelly/rpc_device/device.py
+++ b/aioshelly/rpc_device/device.py
@@ -291,11 +291,11 @@ class RpcDevice:
         if fetch_dynamic := self._supports_dynamic_components():
             calls.append(("Shelly.GetComponents", {"dynamic_only": True}))
         results = await self.call_rpc_multiple(calls)
-        if fetch_dynamic:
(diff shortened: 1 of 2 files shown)


aioshelly/rpc_device/device.py
```diff
@@ -197,23 +197,12 @@ async def _connect_websocket(self) -> None:
...
-                self.aiohttp_session,
-                self.options.ip_address,
-                self.options.device_mac,
-                self.options.port,
-            )
-
-            if self.requires_auth:
-                if self.options.username is None or self.options.password is None:
-                    raise InvalidAuthError("auth missing and required")  # noqa: TRY301
-
```
bdraco, line 210: This should still raise from the WebSocket instead

bdraco:
I don't think this will work

```
2024-08-27 10:13:47.295 ERROR (MainThread) [homeassistant.components.shelly] Unexpected exception
  File "/Users/bdraco/home-assistant/homeassistant/components/shelly/config_flow.py", line 216, in async_step_credentials
    device_info = await validate_input(
                  ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/bdraco/home-assistant/homeassistant/components/shelly/config_flow.py", line 109, in validate_input
    await rpc_device.initialize()
  File "/Users/bdraco/home-assistant/venv/lib/python3.12/site-packages/aioshelly/rpc_device/device.py", line 189, in initialize
    await self._connect_websocket()
  File "/Users/bdraco/home-assistant/venv/lib/python3.12/site-packages/aioshelly/rpc_device/device.py", line 202, in _connect_websocket
    self.shelly["auth_domain"],
    ^^^^^^^^^^^
  File "/Users/bdraco/home-assistant/venv/lib/python3.12/site-packages/aioshelly/rpc_device/device.py", line 433, in shelly
    raise NotInitialized
aioshelly.exceptions.NotInitialized
```

aioshelly/rpc_device/device.py
```diff
@@ -285,6 +268,18 @@ async def update_config(self) -> None:
 
     async def _init_calls(self) -> None:
         """Make calls needed to initialize the device."""
+        # _supports_dynamic_components() needs _shelly to be set
+        # so we need to fetch it first. In the future we could
+        # change this to fetch all data in one call once call_rpc_multiple
+        # can be refactored to return errors instead of raising them.
+        self._shelly = await self.call_rpc("Shelly.GetDeviceInfo")
```
bdraco, line 275: GetDeviceInfo does not require auth per the docs

aioshelly/rpc_device/device.py
```diff
@@ -285,6 +268,18 @@ async def update_config(self) -> None:
 
     async def _init_calls(self) -> None:
         """Make calls needed to initialize the device."""
+        # _supports_dynamic_components() needs _shelly to be set
+        # so we need to fetch it first. In the future we could
+        # change this to fetch all data in one call once call_rpc_multiple
+        # can be refactored to return errors instead of raising them.
```
bdraco, line 274:
```suggestion
        # Shelly.GetDeviceInfo is the only RPC call that does not
        # require auth, so we must do a separate call here to get
        # the auth_domain/id
```

Adjusted comment as we need to keep it separate

aioshelly/rpc_device/device.py
```diff
@@ -285,6 +268,17 @@ async def update_config(self) -> None:
 
     async def _init_calls(self) -> None:
         """Make calls needed to initialize the device."""
+        # Shelly.GetDeviceInfo is the only RPC call that does not
+        # require auth, so we must do a separate call here to get
+        # the auth_domain/id
```
bdraco, line 273:
```suggestion
        # the auth_domain/id so we can enable auth for the rest of the calls
```

thecode (approved):
Nice, tested with Shelly Plus HT and Shelly Plus 1PM Gen3 with and without auth

bdraco:
Did some more testing. All seems good.

PR #640, opened 2024-08-27
Use `device_id` as `realm` value for auth

bieniu:
This will fix WallDisplay support with authorization enabled. WallDisplay doesn't return `auth_domain` value in response to `/shelly`.


chemelli74:
Tested with 3 devices and working!

thecode (approved):

PR #643, opened 2024-08-30
Add a separate timeout error exception class

bdraco:
When `DeviceConnectionError` is raised, its not immediately obvious that when its a timeout problem.

`DeviceConnectionTimeoutError` is a subclass of `DeviceConnectionError` to ensure backwards compat.

related issue https://github.com/home-assistant/core/issues/124888

### DIFF
--- a/aioshelly/block_device/device.py
+++ b/aioshelly/block_device/device.py
@@ -186,7 +186,7 @@ class BlockDevice:
             raise
         except TimeoutError as err:
             self._last_error = DeviceConnectionTimeoutError(err)
-            _LOGGER.debug("host %s: error: %r", ip, self._last_error)
+            _LOGGER.debug("host %s: timeout error: %r", ip, self._last_error)
             raise self._last_error from err
         except CONNECT_ERRORS as err:
             self._last_error = DeviceConnectionError(err)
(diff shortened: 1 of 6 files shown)


aioshelly/block_device/device.py
```diff
@@ -183,6 +184,10 @@ async def initialize(self) -> None:
             self._last_error = err
             _LOGGER.debug("host %s: error: %r", ip, err)
             raise
+        except TimeoutError as err:
+            self._last_error = DeviceConnectionTimeoutError(err)
+            _LOGGER.debug("host %s: error: %r", ip, self._last_error)
```
bdraco, line 189:
```suggestion
            _LOGGER.debug("host %s: timeout error: %r", ip, self._last_error)
```

chemelli74 (approved):
LGTM!

thecode:
Although changes are backward compatible, I think it would be better to update the tools https://github.com/home-assistant-libs/aioshelly/blob/ddf54d473e9bda557df7d48660f5f2b66bc41cdb/tools/common/__init__.py#L69-L70 since we use them to test/debug PRs and would be nice to show that the error is a timeout and not a connection problem

bieniu (approved):
Looks good 👍

PR #649, opened 2024-09-01
Fix treating null RPC responses as failures

bdraco:


tests/rpc_device/test_wsrpc.py
```diff
@@ -0,0 +1,48 @@
+"""Tests for rpc_device.device module."""
```
bdraco, line 1:
```suggestion
"""Tests for rpc_device.wsrpc module."""
```

thecode (approved):
Reproduced the problem with Shelly Plus 2PM in cover mode and verified that the PR fix it

PR #658, opened 2024-09-04
Add tests for `get_info()`

bieniu:


chemelli74:
I would prefer to use the fixtures we have as the only purpose was to use them for tests.

thecode:
> I would prefer to use the fixtures we have as the only purpose was to use them for tests.

If you look at <https://github.com/home-assistant-libs/aioshelly/tree/main/tests/rpc_device/fixtures> it already contain some specific fixtures, I think for testing specific methods we don't need to use a full device fixture, also a full device fixture will not fit when you want to mock an error response.
As I understand the full device fixtures will be used when we have a device simulator which will simulate the behavior of the device but we are not there yet.

bieniu:
I agree with Shay, we can use device fixtures to simulate the entire device. Here we are testing one small function.

thecode (approved):

PR #688, opened 2024-10-15
feat: add config and status for BLU TRV

chemelli74:


aioshelly/rpc_device/device.py
```diff
@@ -304,7 +305,9 @@ async def _init_calls(self) -> None:
         if fetch_status:
             self._status = results.pop(0)
         if fetch_dynamic:
-            self._parse_dynamic_components(results.pop(0))
+            components = results.pop(0)
+            self._parse_dynamic_components(components)
+            await self._retrieve_blutrv_components(components)
```
bieniu, line 310: I think we should not call this for all devices that support dynamic components but only for BLU Gateway G3.

chemelli74: Good point, will update code

aioshelly/rpc_device/device.py
```diff
@@ -533,3 +537,26 @@ def _parse_dynamic_components(self, components: dict[str, Any]) -> None:
                 for item in self._dynamic_components
             }
         )
+
+    async def _retrieve_blutrv_components(self, components: dict[str, Any]) -> None:
```
thecode, line 548:
```suggestion
    async def _retrieve_blu_components(self, components: dict[str, Any]) -> None:
```

I think it would be better not to tie this only with TRV so we are ready to support other BLU devices in the future

aioshelly/rpc_device/device.py
```diff
@@ -538,3 +544,29 @@ def _parse_dynamic_components(self, components: dict[str, Any]) -> None:
...
+        """Retrieve BLU TRV components."""
+        if self.model != MODEL_BLU_GATEWAY_GEN3:
+            return
+
+        if TYPE_CHECKING:
+            assert self._config
+            assert self._status
```
thecode, line 555: Since we can call this method when periodically pulling a device and we had history of race conditions of initializing the device multiple times, I think we should explicitly check that these values are not none and log error if they are.

aioshelly/rpc_device/device.py
chemelli74: Logic so far is quite tie to the `blutrv`. We can always update it once we have new devices.
After all is an internal method, not a public one.

aioshelly/rpc_device/device.py
```diff
@@ -538,3 +544,28 @@ def _parse_dynamic_components(self, components: dict[str, Any]) -> None:
...
+                    ("BluTrv.GetRemoteConfig", {"id": _key[1]}),
+                    ("BluTrv.GetRemoteStatus", {"id": _key[1]}),
+                ]
+                results = await self.call_rpc_multiple(calls)
+
+                cfg: dict[str, Any] = results[0]["config"]["trv:0"]
+                # addr, name and model_id must be added from Shelly.GetComponents call
+                cfg.update({"addr": component["config"]["addr"]})
+                cfg.update({"name": component["config"]["name"]})
+                cfg.update({"model_id": component["attrs"].get("model_id")})
```
chemelli74, line 569: Once #737 is merged, I'll change it to the looked up value

aioshelly/rpc_device/device.py
```diff
@@ -538,3 +544,28 @@ def _parse_dynamic_components(self, components: dict[str, Any]) -> None:
...
             }
         )
+
+    async def _retrieve_blutrv_components(self, components: dict[str, Any]) -> None:
+        """Retrieve BLU TRV components."""
+        if self.model != MODEL_BLU_GATEWAY_GEN3:
+            return
+
+        if not self._config or not self._status:
+            raise ConnectionError
```
thecode, line 554:
```suggestion
            raise NotInitialized
```
We don't have any `ConnectionError` handling in `aioshelly`

aioshelly/rpc_device/device.py
```diff
@@ -538,3 +544,28 @@ def _parse_dynamic_components(self, components: dict[str, Any]) -> None:
...
+        for component in components.get("components", []):
+            _key = component["key"].split(":")
+            if _key[0] == BLU_TRV_IDENTIFIER:
+                calls = [
+                    ("BluTrv.GetRemoteConfig", {"id": _key[1]}),
+                    ("BluTrv.GetRemoteStatus", {"id": _key[1]}),
+                ]
+                results = await self.call_rpc_multiple(calls)
+
+                cfg: dict[str, Any] = results[0]["config"]["trv:0"]
```
thecode, line 565: What happens if there is more than 1 TRV?

aioshelly/rpc_device/device.py
```diff
@@ -538,3 +544,28 @@ def _parse_dynamic_components(self, components: dict[str, Any]) -> None:
...
+                # addr, name and model_id must be added from Shelly.GetComponents call
+                cfg.update({"addr": component["config"]["addr"]})
+                cfg.update({"name": component["config"]["name"]})
+                cfg.update({"model_id": component["attrs"].get("model_id")})
+                self._config.update({component["key"]: cfg})
+                self._status.update({component["key"]: results[1]["status"]["trv:0"]})
```
thecode, line 571: What happens if there is more than 1 TRV?

thecode (changes requested):

aioshelly/rpc_device/device.py
chemelli74: The id of the TRV changes (200,201,202) , so the key changes

aioshelly/rpc_device/device.py
chemelli74: Same answer as above. Here we are already inside the "blutrv:200" key.
For the second one it will be "blutrv:201", etc.

aioshelly/rpc_device/device.py
```diff
@@ -538,3 +545,34 @@ def _parse_dynamic_components(self, components: dict[str, Any]) -> None:
...
+                results = await self.call_rpc_multiple(calls)
+
+                cfg: dict[str, Any] = results[0]["config"]["trv:0"]
+                # addr, name and model_id must be added from Shelly.GetComponents call
+                cfg.update({"addr": component["config"]["addr"]})
+                cfg.update({"name": component["config"]["name"]})
+                cfg.update(
+                    {
+                        "model_id": BLU_TRV_MODEL_ID.get(
+                            component["attrs"].get("model_id")
```
bieniu, line 573: I get `KeyError` here:
```
  File "/home/maciek/develop/aioshelly/aioshelly/rpc_device/device.py", line 573, in _retrieve_blutrv_components
    component["attrs"].get("model_id")
    ~~~~~~~~~^^^^^^^^^
KeyError: 'attrs'
```

aioshelly/rpc_device/device.py
```diff
@@ -538,3 +545,34 @@ def _parse_dynamic_components(self, components: dict[str, Any]) -> None:
...
+        """Retrieve BLU TRV components."""
+        if self.model != MODEL_BLU_GATEWAY_GEN3:
+            return
+
+        if not self._config or not self._status:
+            raise NotInitialized
+
+        for component in components.get("components", []):
+            _key = component["key"].split(":")
+            if _key[0] == BLU_TRV_IDENTIFIER:
```
bieniu, line 559: Maybe?
```suggestion
            if _key[0] != BLU_TRV_IDENTIFIER:
                continue
```

aioshelly/rpc_device/device.py
bieniu: `attr.model_id` is under `bthomedevice:200`, not `blutrv:200`

obraz

chemelli74: As you know from Teams, they are going to add it to the right key.
But I was under the impression that attrs was already there.

We can do a get for that key as well or wait for the new firmware

bieniu: We don't know when the Shelly team will deliver such a change and whether it will be part of the beta firmware that will be available for users. I think for now we should use `.get()` here to make sure our code doesn't return an exception.

chemelli74: They wrote in Teams: "ok, I'll try to make the addition tomorrow and also send you the table."

Anyway, will switch to get

aioshelly/rpc_device/device.py
thecode: I am asking about the `trv:0`, do we know it will always be `trv:0`?

aioshelly/rpc_device/device.py
chemelli74: Why ? We need to make the call only for "blutrv" so only if the key is present

bieniu: Less indentation for subsequent lines

aioshelly/rpc_device/device.py
chemelli74: Yeah I got your question and yes, there will be always

thecode: I am not sure, it is odd to hardcode `trv:0` and I fear that with more than one TRV this may get another id (`trv:1`), even if they are on a different key, I think it worth checking with the Shelly team just to be sure.

chemelli74: Confirmed by Shelly, trv:0 will always have id zero

aioshelly/const.py
```diff
@@ -144,10 +144,13 @@
 # Firmware 1.0.99 release date
 GEN3_MIN_FIRMWARE_DATE = 20231102
 
+# Firmware 1.5.x release date
+# Temporary use beat release to allow BluTrv support
```
bieniu, line 148:
```suggestion
# Temporary use beta release to allow BluTrv support
```

thecode (approved):

aioshelly/rpc_device/device.py
```diff
@@ -568,10 +568,15 @@ async def _retrieve_blutrv_components(self, components: dict[str, Any]) -> None:
 
             cfg: dict[str, Any] = results[0]["config"]["trv:0"]
             # addr, name and model_id must be added from Shelly.GetComponents call
+            _attrs = component.get("attrs")
```
thecode, line 571:
```suggestion
            _attrs = component.get("attrs", {})
```

aioshelly/rpc_device/device.py
```diff
@@ -568,10 +568,15 @@ async def _retrieve_blutrv_components(self, components: dict[str, Any]) -> None:
...
             cfg.update({"addr": component["config"]["addr"]})
             cfg.update({"name": component["config"]["name"]})
             cfg.update(
-                {"local_name": BLU_TRV_MODEL_ID.get(component["attrs"]["model_id"])}
+                {
+                    "local_name": BLU_TRV_MODEL_ID.get(_attrs.get("model_id"))
+                    if _attrs
+                    else None
+                }
```
thecode, line 579: If we set the default to `{}` we don't need this condition

chemelli74: Good catch !

aioshelly/const.py
```diff
@@ -144,10 +144,13 @@
 # Firmware 1.0.99 release date
 GEN3_MIN_FIRMWARE_DATE = 20231102
 
+# Firmware 1.5.x release date
+# Temporary use beta release to allow BluTrv support
+GEN3_GATEWAY_MIN_FIRMWARE_DATE = 20241219
```
bieniu, line 149: Firmware `20250109-090619/1.5.0-beta1-g16e0ef4` is available.
```suggestion
GEN3_GATEWAY_MIN_FIRMWARE_DATE = 20250109
```

bieniu (approved):

thecode (approved):

PR #689, opened 2024-10-15
feat: add Shelly BLU Gateway GEN3 model

chemelli74:


aioshelly/const.py
```diff
@@ -460,6 +461,13 @@ class ShellyDevice:
         gen=GEN2,
         supported=True,
     ),
+    MODEL_BLU_GATEWAY_GEN3: ShellyDevice(
+        model="S3GW-1DBT001",
+        name="Shelly BLU Gateway GEN3",
```
thecode, line 466:
```suggestion
        name="Shelly BLU Gateway Gen3",
```
So it match how it written on the box and other similar models in this file

thecode (changes requested):

thecode (approved):

PR #715, opened 2024-11-16
Fix MacAddressMismatchError check for RPC devices

thecode:
Fixes regression found at https://github.com/home-assistant/core/issues/130696


bieniu (approved):
Looks good, great catch!

thecode:
Thanks

Issue #730, opened 2024-12-02
Shelly Pro 1PM is an unkown Modell in Shelly Integration


lgusewski:
I [tried](https://github.com/home-assistant/core/issues/131959) to inform about the issue, that my Shelly Pro 1PM is an unknown device in Home Assistant 2024.11.3 with the official Shelly integration. But the maintainer informed me, that this must be a bug in the aioshelly library. So please check, my device shows "SPSW-201PE15UL" as device info and has the firmware 20241011-114451/1.4.4-g6d2a586. It is gen2.

bieniu:
Please show us a response to: `http://<DEVICE_IP>/shelly`

lgusewski:
{"name":null,"id":"shellypro1pm-************","mac":"************","slot":1,"model":"SPSW-201PE15UL","gen":2,"fw_id":"20241011-114451/1.4.4-g6d2a586","ver":"1.4.4","app":"Pro1PM","auth_en":false,"auth_domain":null}

PR #731, opened 2024-12-02
Add model ID for Pro 1PM UL (SPSW-201PE15UL)

bieniu:
Closes https://github.com/home-assistant-libs/aioshelly/issues/730


chemelli74 (approved):
Thx!

PR #750, opened 2025-01-06
Add info about Gen4 devices

bieniu:


aioshelly/const.py
```diff
@@ -918,6 +922,34 @@ class ShellyDevice:
         gen=GEN3,
         supported=True,
     ),
+    MODEL_1PM_G4: ShellyDevice(
+        model="S4SW-001P16EU",
```
chemelli74, line 926: Why we always write the model manually instead of using MODEL_1PM_G4 as it's already there ?

bieniu: Good question, I don't know 😄

chemelli74:
Thx!

chemelli74 (approved):
LGTM

thecode (approved):

PR #769, opened 2025-01-27
Remove ruff rule COM812 from ignore list

bieniu:


thecode:
I am not sure if we should enable this rule, this is the warning when it is enabled:
`warning: The following rule may cause conflicts when used with the formatter: `COM812`. To avoid unexpected behavior, we recommend disabling this rule, either by removing it from the `select` or `extend-select` configuration, or adding it to the `ignore` configuration.`

The docs recommend to disable this rule if `ruff-format` is used:
https://docs.astral.sh/ruff/rules/missing-trailing-comma/
"We recommend against using this rule alongside the [formatter](https://docs.astral.sh/ruff/formatter/). The formatter enforces consistent use of trailing commas, making the rule redundant."

We use ruff-format:
https://github.com/home-assistant-libs/aioshelly/blob/82beda6d97876d42a1ef4c7210c424af21a39640/.pre-commit-config.yaml#L13

It is also redundant once `ruff-format` is used as it would add trailing comma where needed.
For `aiowebostv` I added a comment why it is disabled so I don't get back to it:
https://github.com/home-assistant-libs/aiowebostv/blob/dba108947211b16a90259733b971ff0d0d11653d/ruff.toml#L8

bieniu:
Oh yes, you're right, I completely forgot about this. I'm closing PR.

PR #770, opened 2025-01-28
Add notes for ruff ignored rules that conflicts with other rules

thecode:
Add notes for ruff ignored rules that conflicts with other rules so we don't get back to them every time, we try to cleanup ignored rules, related to https://github.com/home-assistant-libs/aioshelly/pull/769


bieniu (approved):
Thanks!

PR #775, opened 2025-01-31
Add xmod_info() method for XMOD devices

chemelli74:


aioshelly/rpc_device/device.py
```diff
@@ -498,7 +501,10 @@ def version(self) -> str:
     @property
     def model(self) -> str:
         """Device model."""
-        return cast(str, self.shelly["model"])
+        model = cast(str, self.shelly["model"])
+        if model in SHELLY_X_MODELS:
+            return f"{model}-{self.shelly['jwt']['p']}"
```
thecode, line 506: I don't think we should mix two models from 2 vendors as a single model name.
The model property should return what we get from `shelly["model"].
We can add another property that will return the vendor model, but I don't see the benefit of it as the `jwt` already contains the model name so we are just duplicating data into `SHELLY_X_MODELS`

thecode (dismissed):

aioshelly/rpc_device/device.py
```diff
@@ -338,6 +339,8 @@ async def get_all_pages(self, first_page: dict[str, Any]) -> dict[str, Any]:
 
     async def script_list(self) -> list[ShellyScript]:
         """Get a list of scripts from 'Script.List'."""
+        if self.model == MODEL_X_BASE:
+            return []
```
thecode, line 343: We don't call `script_list` anywhere during init in `aioshelly` we call it from Shelly integration when we setup Bluetooth, we should completely disable the usage of Bluetooth same as we do for sleeping devices and for the wall display:
https://github.com/home-assistant/core/blob/1a394876b11c85893998595425244e84f157abe8/homeassistant/components/shelly/config_flow.py#L458-L464

thecode (dismissed):

aioshelly/rpc_device/device.py
```diff
@@ -26,6 +26,7 @@
     DEVICE_POLL_TIMEOUT,
     FIRMWARE_PATTERN,
     MODEL_BLU_GATEWAY_GEN3,
+    MODEL_X_BASE,
```
thecode, line 29:
```suggestion
```

aioshelly/rpc_device/device.py
thecode, line 343:
```suggestion
```

aioshelly/rpc_device/device.py
chemelli74: I think we can keep it to use on HA side to filter Bluetooth feature for this class of devices

thecode: I think that we are assuming that all X models will have this part number, but this is not correct, we are also missing a `ShellyDevice` for this added model (not for the `jwt` model, for the base model)

thecode (approved):

bieniu (approved):

PR #779, opened 2025-02-03
Remove XMOD model 

chemelli74:
For XMOD models product and name comes from JWT block and not from the standard "model" key

### DIFF
--- a/aioshelly/const.py
+++ b/aioshelly/const.py
@@ -123,8 +123,6 @@ MODEL_I4_GEN3 = "S3SN-0024X"
 MODEL_OUT_PLUG_S_G3 = "S3PL-20112EU"
 MODEL_PM_MINI_G3 = "S3PM-001PCEU16"
 MODEL_PLUG_S_G3 = "S3PL-00112EU"
-MODEL_X_MOD1 = "S3MX-0A"
-MODEL_X_MOD_S3XT_0S = "S3XT-0S"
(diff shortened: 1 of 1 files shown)


aioshelly/const.py
```diff
@@ -123,7 +123,6 @@
 MODEL_OUT_PLUG_S_G3 = "S3PL-20112EU"
 MODEL_PM_MINI_G3 = "S3PM-001PCEU16"
 MODEL_PLUG_S_G3 = "S3PL-00112EU"
-MODEL_X_MOD1 = "S3MX-0A"
```
bieniu, line 126: What about X devices without configuration from x.shelly.cloud? JWT block is not available for them.

```
{"name":null,"id":"shellyxmod1-xxxxxxxx","mac":"xxxxxxxx","slot":0,"model":"S3MX-0A","gen":3,"fw_id":"20241011-114452/1.4.4-g6d2a586","ver":"1.4.4","app":"XMOD1","auth_en":false,"auth_domain":null,"jti":"000000000000"}
```

chemelli74: I though that all X devices need a  X.shelly.cloud config, but seems not.
Let me ask Shelly for a clarification.

Thx for spotting.

chemelli74:
Devkit doesn't need provisioning via x.shelly.cloud

PR #780, opened 2025-02-04
Clean up MODEL constants

bieniu:
A little cleaning for the model consts:
 - use `MODEL_*` const as `ShellyDevice.model`
 - use `Gx` suffix in `MODEL_*` const names
 - sort consts
 
I think we should mark this as a breaking change. The `MODEL_BLU_GATEWAY_GEN3` constant is used in HA tests, we need to update its name when bumping `aioshelly`.


chemelli74 (approved):
LGTM !

thecode (approved):

PR #782, opened 2025-02-06
Update status instead of overwriting it when pooling

bieniu:
Fixes `KeyError` when `rssi` and `uptime` sensors are enable for BLU Gateway

```
2025-02-06 15:13:59.755 DEBUG (MainThread) [aioshelly.rpc_device.wsrpc] send(192.168.2.210:80): {'id': 12, 'method': 'Shelly.GetComponents', 'src': 'aios-140615453547488', 'params': {'dynamic_only': True}, 'dst': 'shellyblugwg3-34cdb078742c', 'auth': {'realm': 'shellyblugwg3-34cdb078742c', 'username': 'admin', 'nonce': 1738849439, 'cnonce': 1738851239, 'response': '6849f9f88eef4ebb4c22c0e4d1948a1c371432d137688b75232c1097ec614a22', 'algorithm': 'SHA-256'}}
2025-02-06 15:13:59.760 ERROR (MainThread) [homeassistant] Error doing job: Task exception was never retrieved (None)
  File "/home/maciek/develop/home-assistant-core/venv/lib/python3.13/site-packages/aioshelly/rpc_device/wsrpc.py", line 371, in _rx_msgs
    self.handle_frame(RPCSource.CLIENT, frame)
    ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/maciek/develop/home-assistant-core/venv/lib/python3.13/site-packages/aioshelly/rpc_device/wsrpc.py", line 315, in handle_frame
    self._on_notification(source, method, params)
    ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/maciek/develop/home-assistant-core/venv/lib/python3.13/site-packages/aioshelly/rpc_device/device.py", line 171, in _on_notification
    self._update_listener(self, update_type)
    ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^
  File "/home/maciek/develop/home-assistant-core/homeassistant/components/shelly/coordinator.py", line 754, in _async_handle_update
    self.async_set_updated_data(None)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
  File "/home/maciek/develop/home-assistant-core/homeassistant/helpers/update_coordinator.py", line 515, in async_set_updated_data
    self.async_update_listeners()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/home/maciek/develop/home-assistant-core/homeassistant/helpers/update_coordinator.py", line 178, in async_update_listeners
    update_callback()
    ~~~~~~~~~~~~~~~^^
  File "/home/maciek/develop/home-assistant-core/homeassistant/components/shelly/entity.py", line 391, in _update_callback
    self.async_write_ha_state()
    ~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/home/maciek/develop/home-assistant-core/homeassistant/helpers/entity.py", line 1023, in async_write_ha_state
    self._async_write_ha_state()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/home/maciek/develop/home-assistant-core/homeassistant/helpers/entity.py", line 1148, in _async_write_ha_state
    self.__async_calculate_state()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/home/maciek/develop/home-assistant-core/homeassistant/helpers/entity.py", line 1085, in __async_calculate_state
    state = self._stringify_state(available)
  File "/home/maciek/develop/home-assistant-core/homeassistant/helpers/entity.py", line 1029, in _stringify_state
    if (state := self.state) is None:
                 ^^^^^^^^^^
  File "/home/maciek/develop/home-assistant-core/homeassistant/components/binary_sensor/__init__.py", line 200, in state
    if (is_on := self.is_on) is None:
                 ^^^^^^^^^^
  File "/home/maciek/develop/home-assistant-core/homeassistant/components/shelly/binary_sensor.py", line 71, in is_on
    return bool(self.attribute_value)
                ^^^^^^^^^^^^^^^^^^^^
  File "/home/maciek/develop/home-assistant-core/homeassistant/components/shelly/entity.py", line 564, in attribute_value
    self.status.get(self.entity_description.sub_key), self._last_value
    ^^^^^^^^^^^
  File "/home/maciek/develop/home-assistant-core/homeassistant/components/shelly/entity.py", line 381, in status
    return cast(dict, self.coordinator.device.status[self.key])
                      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^
KeyError: 'blutrv:200'
```

### DIFF
--- a/aioshelly/rpc_device/device.py
+++ b/aioshelly/rpc_device/device.py
@@ -284,7 +284,7 @@ class RpcDevice:
         results = await self.call_rpc_multiple(calls, DEVICE_POLL_TIMEOUT)
         if (status := results[0]) is None:
             raise RpcCallError("empty response to Shelly.GetStatus")
-        if not self._status:
+        if self._status is None:
             raise NotInitialized
         self._status.update(status)
         if has_dynamic:
--- a/aioshelly/rpc_device/device.py
+++ b/aioshelly/rpc_device/device.py
@@ -284,7 +284,9 @@ class RpcDevice:
         results = await self.call_rpc_multiple(calls, DEVICE_POLL_TIMEOUT)
         if (status := results[0]) is None:
             raise RpcCallError("empty response to Shelly.GetStatus")
-        self._status = status
+        if not self._status:
+            raise NotInitialized
+        self._status.update(status)
         if has_dynamic:
             if (dynamic := results[1]) is None:
                 raise RpcCallError("empty response to Shelly.GetComponents")


bieniu:
Relates to https://github.com/home-assistant/core/issues/137575

bdraco:
I don't think the coordinator is polling in this case and its only updating with manual data `async_set_updated_data` via the `_on_notification` callback which is called in the `rx_msgs` loop in `aioshelly`. It looks like we are missing a guard in <https://github.com/home-assistant-libs/aioshelly/blob/b6dd433232b9f9ff30af6ed1f6e9b25bae8b7418/aioshelly/rpc_device/wsrpc.py#L315> to check if the callback throws and exception call `_LOGGER.exception` there so it ends up all the way up the asyncio stack.

thecode:
> I don't think the coordinator is polling in this case and its only updating with manual data `async_set_updated_data` via the `_on_notification` callback which is called in the `_rx_msgs` loop in `aioshelly`. It looks like we are missing a guard in
> 

We have two coordinators, one is polling and one push data, I understand that this starts to happen after enabling one of the polling entities (until you enable polling entities the polling coordinator doesn't do anything). I don't think there is a race here, I think there are cases that the `blutrv` key is not reported when we poll status, might be that the gateway is trying to get an updated data from the device and reply without this data. In any case it is safer not to delete the existing data. We do similar for the push coordinator as we learned some push messages are partial and when we overwrite the whole status we miss data and break entities.

bdraco:
Sorry for the lack of clarity, I think this change is fine.

I'm concerned that we shouldn't let the callback exception get all the way up the asyncio stack so it gets caught by the loop exception handler.

PR #790, opened 2025-02-10
Models exception list for BT Proxy support

chemelli74:


aioshelly/const.py
```diff
@@ -1019,6 +1019,8 @@ class UndefinedType(Enum):
 
 MODEL_NAMES = {data.model: data.name for data in DEVICES.values()}
 
+MODELS_NOT_SUPPORTING_BT_PROXY = [MODEL_WALL_DISPLAY]
```
bdraco, line 1022: Since we call it `async_script_supported ` on the HA side https://github.com/home-assistant/core/pull/137323/files#diff-905d57e5ed1df19aec0a365e68cf5ad1426da82ebe9199899e25097ab90ee5abR85 I think it makes sense to name this for models that do not have working script support instead

chemelli74:
Superseded by #800

PR #792, opened 2025-02-11
Add BLE support check

thecode:
Detect if device supports BLE by trying to read a script from the device instead of hardcoded models list and catch known errors.
The method checks for BLE support although for now it is only breaks if scripts are not supported so we can expand it in the future if another device has a script but does not support BLE scanning.

Updated the example to be able to test the new method on real devices.
Tested on:
- Shelly Wall Display --> return `False`
- Shelly Wall Display X2 --> return `False`
- LinkedGo Starlight Thermostat ST1820 --> return `False`
- Shelly Plus 2PM with existing script --> return `True`
- Shelly Pro 2 without script --> return `True`

### DIFF
--- a/aioshelly/ble/__init__.py
+++ b/aioshelly/ble/__init__.py
@@ -12,6 +12,9 @@ from .backend.scanner import ShellyBLEScanner
 from .const import (
     BLE_CODE,
     BLE_SCRIPT_NAME,
+    RPC_CALL_ERR_INVALID_ARG,
+    RPC_CALL_ERR_METHOD_NOT_FOUND,
+    RPC_CALL_ERR_NO_HANDLER,
     VAR_ACTIVE,
     VAR_EVENT_TYPE,
     VAR_VERSION,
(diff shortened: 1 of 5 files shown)


bieniu (approved):
Looks good 👍

aioshelly/ble/__init__.py
```diff
@@ -105,3 +108,37 @@ async def async_ensure_ble_enabled(device: RpcDevice) -> bool:
     LOGGER.info("BLE enabled, restarting device %s:%s", device.ip_address, device.port)
     await device.trigger_reboot(3500)
     return True
+
+
+async def async_ble_supported(device: RpcDevice) -> bool:
+    """Check if BLE is supported.
```
bdraco, line 114: I think its a bit confusing to name it `ble_supported` as it doesn't show the intent of the function, which is checking if scripts are supported

thecode:
Superseded by https://github.com/home-assistant-libs/aioshelly/pull/800

PR #795, opened 2025-02-15
Move mypy and ruff configuration to pyproject.toml file

bieniu:
In the future we may also move metadata from setup.py to pyproject.toml.


pyproject.toml
```diff
@@ -1,3 +1,27 @@
+[tool.pytest.ini_options]
+asyncio_default_fixture_loop_scope = "function"
```
bieniu, line 2: Fixes:

PytestDeprecationWarning: The configuration option "asyncio_default_fixture_loop_scope" is unset.
The event loop scope for asynchronous fixtures will default to the fixture caching scope. Future versions of pytest-asyncio will default the loop scope for asynchronous fixtures to function scope. Set the default fixture loop scope explicitly in order to avoid unexpected behavior in the future. Valid fixture loop scopes are: "function", "class", "module", "package", "session"

  warnings.warn(PytestDeprecationWarning(_DEFAULT_FIXTURE_LOOP_SCOPE_UNSET))

thecode (approved):

PR #806, opened 2025-02-25
Fix missing semicolon in BLE Script

rkapsi:
This single return statement with the missing semicolon seems inconsistent with the rest of the script and Shelly's scripting language isn't exactly JavaScript. Some older version thereof even said semicolons are mandatory. This PR is therefore adding the missing semicolon.


thecode (approved):
Thanks @rkapsi

PR #817, opened 2025-03-09
Add support for passing the BluetoothScanningMode to the scanner

bdraco:
We want `habluetooth` to be able to make better decisions about routing advertisement data, however we currently don't know if a shelly scanner is in passive or active mode so this is not possible.

### DIFF
--- a/aioshelly/ble/__init__.py
+++ b/aioshelly/ble/__init__.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 
 import logging
 
-from habluetooth import HaBluetoothConnector
+from habluetooth import BluetoothScanningMode, HaBluetoothConnector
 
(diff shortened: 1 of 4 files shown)


aioshelly/ble/__init__.py
```diff
@@ -85,6 +87,8 @@ def create_scanner(
             can_connect=lambda: False,
         ),
         False,
+        requested_mode=requested_mode,
+        current_mode=current_mode,
```
bdraco, line 91: Shelly will always be in sync so requested_mode and current_mode are always the same

thecode (approved):

bieniu (approved):

chemelli74 (approved):

bdraco:
Thanks. I'll do a release shortly

PR #824, opened 2025-03-15
Bump ruff to version 0.11.0

bieniu:


aioshelly/block_device/coap.py
```diff
@@ -191,7 +191,7 @@ def close(self) -> None:
 
     def connection_made(self, transport: asyncio.BaseTransport) -> None:
         """When the socket is set up."""
-        self.transport = cast(asyncio.DatagramTransport, transport)
+        self.transport = cast("asyncio.DatagramTransport", transport)
```
bieniu, line 194: We need to decide whether we use this new notation or add `TC006` to the ignored rules. Personally, I have no opinion, I like both versions.

`TC006 [*] Add quotes to type expression in `typing.cast()``

thecode: I am in favor of disabling this rule:
1. As you noted it breaks following the type in VScode.
2. It is also disabled in [Home Assistant Core](https://github.com/home-assistant/core/blob/940625505f56f0adef0b0978927d5072ab139250/pyproject.toml#L839)
3. Per [ruff TC006](https://docs.astral.sh/ruff/rules/runtime-cast-value/) this is mainly to enforce constancy since in due to type checking sometimes you have to use quotes but we don't have any in this project anyhow.

bieniu: Yes, same feeling here.

bieniu:
Something weird happened to my local repo. I'm closing this PR and opening a new one.

bieniu:
Superseded by https://github.com/home-assistant-libs/aioshelly/pull/825

PR #833, opened 2025-03-26
Use release-drafter to generate release tag, add categories for PRs

bieniu:


thecode (approved):
LGTM, we should consider editing our GitHub labels and match them to this PR

thecode:
> > we should consider editing our GitHub labels and match them to this PR
> 
> I've already added new labels

I see now that you added, the following are missing:
- performance
- maintenance

bieniu:
> the following are missing

I added the missing ones.

PR #840, opened 2025-04-05
Update script support check to use invalid script ID for validation

thecode:
Related to https://github.com/home-assistant/core/issues/142190#issuecomment-2779662995, reading a large script fails, so instead read an invalid script.

Tested with:
- Shelly Pro 2PM
- Shelly Wall Display X2
- LinkedGo ST1820


bieniu (approved):
Looks good

PR #841, opened 2025-04-05
Read 0 bytes when checking for script support

thecode:
Better alternative to https://github.com/home-assistant-libs/aioshelly/pull/840, instead of reading the whole script, read only 1 byte from the script, related to https://github.com/home-assistant/core/issues/142190#issuecomment-2779662995.

Tested with:
- Shelly Pro 2PM
- Shelly Wall Display X2
- LinkedGo ST1820

### DIFF
--- a/aioshelly/rpc_device/device.py
+++ b/aioshelly/rpc_device/device.py
@@ -624,8 +624,8 @@ class RpcDevice:
     async def supports_scripts(self) -> bool:
         """Check if the device supports scripts.
 
-        Try to read 1 byte from a script to check if the device supports scripts,
-        if it supports scripts, it should return the script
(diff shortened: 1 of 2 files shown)


aioshelly/rpc_device/device.py
```diff
@@ -630,7 +638,7 @@ async def supports_scripts(self) -> bool:
         {"code":404,"message":"No handler for Script.GetCode"}
         """
         try:
-            await self.script_getcode(1)
+            await self.script_getcode(1, bytes_to_read=1)
```
bieniu, line 641: I think we can even use `0` bytes here.

Response:
```
{"data":"", "left":0}
```

bieniu (approved):

PR #850, opened 2025-04-14
Add initial tests for block_device module

thecode:
Create initial tests for block_device module, this mainly so that code is imported, only two tests added but now we can expand coverage by adding more tests.


bieniu (approved):
Looks good, thanks 👍

PR #857, opened 2025-04-27
Migrate shelly to use habluetooth _async_on_raw_advertisement API

bdraco:
Some libraries need access to the raw advertising data

see https://github.com/home-assistant/core/issues/143769#issuecomment-2833594159


thecode (approved):

bdraco:
thanks

PR #866, opened 2025-05-16
`tools/example.py` Improvements

alexjurkiewicz:
Minor improvements to example code:
1. Make `ctrl-c` work properly
2. Replace some logic with `argparse` functionality

### DIFF
--- a/tools/common/__init__.py
+++ b/tools/common/__init__.py
@@ -192,8 +192,11 @@ async def update_outbound_ws(
 async def wait_for_keyboard_interrupt() -> None:
     """Wait for keyboard interrupt (Ctrl-C)."""
     sig_event = asyncio.Event()
-    signal.signal(signal.SIGINT, lambda _exit_code, _frame: sig_event.set())
+    loop = asyncio.get_running_loop()
+    loop.add_signal_handler(signal.SIGINT, lambda: sig_event.set())
     await sig_event.wait()
+    # In a real program, you may want to be more graceful here
+    sys.exit(1)
 
 
 async def check_rpc_device_supports_scripts(
--- a/tools/example.py
+++ b/tools/example.py
@@ -127,18 +127,40 @@ def get_arguments() -> tuple[argparse.ArgumentParser, argparse.Namespace]:
     parser.add_argument("--username", "-u", type=str, help="Set device username")
     parser.add_argument("--password", "-p", type=str, help="Set device password")
 
-    parser.add_argument(
-        "--gen1", "-g1", action="store_true", help="Force Gen1 (CoAP) device"
+    gen = parser.add_mutually_exclusive_group()
+    gen.add_argument(
+        "--gen1",
+        "-g1",
+        action="store_const",
+        const=1,
+        dest="gen",
+        help="Force Gen1 (CoAP) device",
     )
-    parser.add_argument(
-        "--gen2", "-g2", action="store_true", help="Force Gen 2 (RPC) device"
+    gen.add_argument(
+        "--gen2",
+        "-g2",
+        action="store_const",
+        const=2,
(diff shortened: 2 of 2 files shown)


tools/common/__init__.py
```diff
@@ -192,8 +192,11 @@ async def update_outbound_ws(
 async def wait_for_keyboard_interrupt() -> None:
     """Wait for keyboard interrupt (Ctrl-C)."""
     sig_event = asyncio.Event()
-    signal.signal(signal.SIGINT, lambda _exit_code, _frame: sig_event.set())
+    loop = asyncio.get_running_loop()
+    loop.add_signal_handler(signal.SIGINT, lambda: sig_event.set())
     await sig_event.wait()
+    # In a real program, you may want to be more graceful here
+    sys.exit(1)
```
thecode, line 199: This was not here by intention to allow a clean shutdown which we also use to verify when we use the example:
https://github.com/home-assistant-libs/aioshelly/blob/bf970c5decf5e1125978274d0c8c3d0295921901/tools/example.py#L47-L48
```suggestion
```

thecode (changes requested):

alexjurkiewicz:
Updated

bieniu (approved):
Looks good, thanks!

tools/example.py
```diff
@@ -52,7 +52,8 @@ async def test_devices(init: bool, gen: int | None) -> None:
     """Test multiple devices."""
     options: ConnectionOptions
 
-    with Path.open("devices.json", encoding="utf8") as fp:
+    p = Path("devices.json")
+    with p.open(encoding="utf8") as fp:  # noqa: ASYNC230 -- this is test code
```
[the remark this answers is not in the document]
alexjurkiewicz: Yes, that works. The linter directive would apply to a longer line. Would you prefer that?

tools/common/__init__.py
```diff
@@ -192,7 +192,8 @@ async def update_outbound_ws(
 async def wait_for_keyboard_interrupt() -> None:
     """Wait for keyboard interrupt (Ctrl-C)."""
     sig_event = asyncio.Event()
-    signal.signal(signal.SIGINT, lambda _exit_code, _frame: sig_event.set())
+    loop = asyncio.get_running_loop()
+    loop.add_signal_handler(signal.SIGINT, lambda: sig_event.set())
```
thecode, line 196: I assume you use a UNIX based environment, these methods are only supported on Unix:
https://peps.python.org/pep-3156/#signal-callbacks
image

I have also tested it on Windows and it does not work, I don't think it worth changing this as many users use this to collect data or test devices on Windows.



```suggestion
    signal.signal(signal.SIGINT, lambda _exit_code, _frame: sig_event.set())
```

tools/example.py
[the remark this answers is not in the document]
thecode:
```suggestion
    with Path.open(Path("devices.json"), encoding="utf8") as fp:
```

tools/example.py
```diff
@@ -178,24 +201,10 @@ async def main() -> None:
     await coap_context.initialize(args.coap_port, args.listen_ip_address)
     await ws_context.initialize(args.ws_port, args.ws_api_url)
 
-    if not args.init and not (args.gen1 or args.gen2 or args.gen3 or args.gen4):
+    if not args.init and not args.gen:
```
thecode, line 204: We use the gen1, gen2, ... in other places that needs to be fixed also:
https://github.com/home-assistant-libs/aioshelly/blob/377414fa8c4bba4c3753b50ce63367bc18f33e36/tools/example.py#L212-L214

thecode (dismissed):

thecode:
Thanks 👍

PR #870, opened 2025-05-20
Add new model ID for Pro 4PM

bieniu:
Info from Teams

image


bieniu:
> You are fast 😄

Good timing, I was just turning on my computer when the message appeared on teams ;)

PR #871, opened 2025-05-20
Fix `wifi` structure in the `config` fixture

bieniu:
Should be

image

### DIFF
--- a/tools/fixture.py
+++ b/tools/fixture.py
@@ -209,6 +209,9 @@ def _redact_rpc_data(data: dict[str, Any]) -> dict[str, Any]:
                 real_mac.lower(), REDACTED_VALUES["device_mac_lower"]
             )
 
+    config["wifi"]["ap"]["ssid"] = config["wifi"]["ap"]["ssid"].replace(
+        real_mac, REDACTED_VALUES["device_mac"]
+    )
     if config["wifi"]["sta"]["ssid"]:
         config["wifi"]["sta"]["ssid"] = REDACTED_VALUES["wifi"]
     if config["wifi"]["sta1"]["ssid"]:


tools/fixture.py
```diff
@@ -209,8 +209,10 @@ def _redact_rpc_data(data: dict[str, Any]) -> dict[str, Any]:
                 real_mac.lower(), REDACTED_VALUES["device_mac_lower"]
             )
 
-    if "wifi" in config:
```
chemelli74, line 212: Please consider devices that are connected only via ethernet

thecode (approved):

tools/fixture.py
chemelli74: Was not the case on old firmware; will recheck when I can.

When I created the fixture script, I did a lot of testing and saved some fixtures as well.
Just a couple of examples for which this script worked:
- https://github.com/home-assistant-libs/aioshelly/blob/main/fixtures/gen2_Shelly%20BLU%20Gateway_SNGW-BT01_20240213-140411-1.2.0-gb1b9aa8.json
- https://github.com/home-assistant-libs/aioshelly/blob/main/fixtures/gen2_Shelly%20Wall%20Display_SAWD-0A1XX10EU1_20240115-105444-1.2.8-59063a9a.json

BTW, there is also gen1 to take into account:
- https://github.com/home-assistant-libs/aioshelly/blob/main/fixtures/gen1_Shelly%20Plug%20S_SHPLG-S_20230913-113421-v1.14.0-gcb84623.json

thecode: I have tested it with a device with the oldest firmware I had (from 2024),  tried to disable Wi-Fi, it works as intended, I think we can proceed with this change, as it is not a production code, if we have a problem in the future with this we can easily fix it.

bieniu: Oh, I forgot about this PR. Thanks for testing and confirming.

PR #877, opened 2025-06-03
Improve Block device initialization 

thecode:
Refactor Block device (Gen1) initialization process by changing UDP based transactions:
- For `cit/d` use HTTP instead of UDP, this is based on TCP now and has internal retry logic, this is the most failed stage when looking at logs from users since this packet is large.
- For `cit/s` HTTP is not available so perform 3 retries before we fail, this is based on `DEVICE_IO_TIMEOUT` so should not increase the device init time upon failures.

### DIFF
--- a/aioshelly/block_device/device.py
+++ b/aioshelly/block_device/device.py
@@ -20,6 +20,7 @@ from ..common import (
     process_ip_or_options,
 )
 from ..const import (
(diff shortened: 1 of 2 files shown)


bieniu (approved):
Tested with Shelly Flood and Shelly 2.5. Looks good!

PR #893, opened 2025-07-13
Fix Block device cit/d for older firmware versions

thecode:
**Shelly Gen1 changelog:**
image

Requesting `cit/d` for Block devices introduced at firmware v1.10 (released at 02.03.2021), currently we still support devices with firmware v1.9:
https://github.com/home-assistant-libs/aioshelly/blob/8df2cdea1d0e947d986da43e46418fe433c9f714/aioshelly/const.py#L148-L149
Mainly due to some of them not having a newer firmware, such as Shelly Plug 1 - latest firmware is v.1.9.4, related to https://github.com/home-assistant/core/issues/148607

This PR checks that the firmware at least at v1.10 for using HTTP `cit/d` requests and fallback to `CoAP` `cit/d` for older devices.

Tested with:
- Shelly Plug 1 with firmware v1.9.4 (latest, fallback)
- Shelly 1 with firmware v1.9.4 (fallback)
- Shelly 1 with firmware v1.10.1 (http)
- Shelly 1 with firmware v1.14.0 (latest, http)
- Shelly 2.5 with firmware v1.14.0 (latest, http)

### DIFF
--- a/aioshelly/block_device/device.py
+++ b/aioshelly/block_device/device.py
@@ -20,10 +20,12 @@ from ..common import (
     process_ip_or_options,
 )
 from ..const import (
-    CIT_S_RETRIES,
+    CIT_RETRIES,
     CONNECT_ERRORS,
     DEFAULT_HTTP_PORT,
     DEVICE_IO_TIMEOUT,
+    FIRMWARE_PATTERN,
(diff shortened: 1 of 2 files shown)


bieniu (approved):
Looks good, thanks!

thecode:
Thanks for the review

PR #922, opened 2025-09-02
Add sensors from object structure

chemelli74:
First working draft, need to test with more devices

### DIFF
--- a/aioshelly/rpc_device/object_mapping.py
+++ b/aioshelly/rpc_device/object_mapping.py
@@ -1,7 +1,9 @@
 """Powered by Shelly object structures."""
 
-OBJECT_STRUCTURES = {
-    "S3XT-0S": {
+from typing import Any
+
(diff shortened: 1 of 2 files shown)


aioshelly/rpc_device/object_mapping.py
```diff
@@ -0,0 +1,101 @@
...
+            "key": "duration",
+            "unit": "min",
+        },
+        "zone1": {
+            "name": "Zone 2",
+            "index": 201,
+            "type": "number",
+            "key": "duration",
+            "unit": "min",
+        },
```
thecode, line 20: The names are set from the device and not fixed, here is an example from my device:
```json
      {
        "name": "Grass Sprinklers",
        "duration": 10,
        "water_amount": 10
      },
      {
        "name": null,
        "duration": 10,
        "water_amount": 10
      },
```

I don't understand why we hardcode the type and unit since we don't do this for any other sensor

PR #976, opened 2025-11-04
Add helpers for BLE WiFi scanning and provisioning

bdraco:
This PR adds reusable helpers for provisioning Shelly devices over Bluetooth. Previously, the provisioning logic was only in the CLI tool, making it hard to use in other contexts like Home Assistant.

## What's new

**Three new modules:**
- `aioshelly/ble/manufacturer_data.py` - Parses Shelly BLE advertisement data to check if a device supports RPC-over-BLE
- `aioshelly/ble/provisioning.py` - Provides `async_scan_wifi_networks()` and `async_provision_wifi()` for WiFi setup via BLE
- `aioshelly/zeroconf.py` - Helper for looking up devices by name using zeroconf

**The provisioning helpers handle:**
- Creating and cleaning up the RPC device connection automatically
- 30-second timeout for WiFi scans (they can be slow)
- Proper error handling and resource cleanup even when things fail

## Changes to existing code

Refactored `tools/ble_provision_wifi.py` to use the new helpers instead of duplicating the logic. The tool is now 32 lines shorter and serves as a good example of how to use the API.

## Testing

Added comprehensive test coverage for all three modules with proper pytest fixtures.

This makes it easy to integrate BLE provisioning into Home Assistant's config flow without duplicating code between the library and integration.

### DIFF
--- a/aioshelly/zeroconf.py
+++ b/aioshelly/zeroconf.py
@@ -28,15 +28,16 @@ async def async_lookup_device_by_name(
     LOGGER.debug("Active lookup for: %s", service_name)
     service_info = AsyncServiceInfo("_http._tcp.local.", service_name)
 
-    if await service_info.async_request(aiozc.zeroconf, 5000):
-        addresses = service_info.parsed_addresses(IPVersion.V4Only)
-        if addresses and service_info.port:
-            host = addresses[0]
-            port = service_info.port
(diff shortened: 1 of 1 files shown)


thecode (approved):
Tested with Shelly Presence Gen4

PR #988, opened 2025-11-19
Fix MAC address byte order in BLE manufacturer data parsing

bdraco:
## Issue

When parsing the MAC address from Shelly Gen4 BLE manufacturer data, the bytes were being read in the wrong order. The MAC address is stored in reverse byte order in the manufacturer data, but the parser was reading them sequentially.

### Example:
- Manufacturer data bytes: `70 d6 c2 97 ba cc`
- Previous (incorrect) parsing: `70:D6:C2:97:BA:CC`
- Correct parsing: `CC:BA:97:C2:D6:70` (bytes reversed)
- Zeroconf name: `shelly1minig4-ccba97c2d670` ✓ matches!

This caused a mismatch between the MAC extracted from BLE and the MAC advertised via zeroconf after WiFi provisioning, breaking the handoff from BLE to WiFi setup.

## Fix

Reverse the MAC address bytes when parsing by using `reversed(mac_bytes)` in the manufacturer data parser.

```python
# Before
result["mac"] = ":".join(f"{b:02X}" for b in mac_bytes)

# After
result["mac"] = ":".join(f"{b:02X}" for b in reversed(mac_bytes))
```

## Testing

Updated all tests to expect the corrected byte order. All 18 manufacturer data tests pass.

### DIFF
--- a/aioshelly/ble/manufacturer_data.py
+++ b/aioshelly/ble/manufacturer_data.py
@@ -58,12 +58,12 @@ def parse_shelly_manufacturer_data(
             offset += 2
 
         elif block_type == BLOCK_TYPE_MAC:
-            # 6 bytes MAC address
+            # 6 bytes MAC address (stored in reverse order)
             if offset + 6 > len(data):
                 break
             mac_bytes = data[offset : offset + 6]
-            # Format as standard MAC address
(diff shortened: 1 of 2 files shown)


bdraco:
Thanks. I'll do a patch release and bump HA

PR #998, opened 2025-11-24
Add model_id for gen2 US plug

bdraco:
I couldn't find this in the shelly docs, but I was able to extract it from the adv data


bdraco:
Thanks

bdraco:
Screenshot 2025-11-23 at 5 52 23 PM

Works now.

I was also testing the other PR to make sure they still come up before the started event when discovery runs

and than
Screenshot 2025-11-24 at 2 18 32 PM


Than I just get

Screenshot 2025-11-24 at 2 19 31 PM

PR #1010, opened 2025-12-01
Ensure RPC is not enabled during BLE activation

thecode:
* Changed the `async_ensure_ble_enabled` function to set `enable_rpc=False` when enabling BLE, ensuring RPC is not unintentionally enabled during BLE activation.

* Added tests in `tests/ble/test_init.py` to verify `async_ensure_ble_enabled` behavior for already enabled BLE, and for scenarios requiring device restart, increasing coverage and robustness.


thecode:
Thanks!

PR #1013, opened 2025-12-04
Add new BLE models and remove duplicates

thecode:
- Add new BLE model IDs received from Shelly

- Remove duplicates and add test to prevent adding duplicate models

If we don't remove duplicates when we fetch the device name by model ID we can get non predictive/incorrect model name:
**Before:**
 ```python
>>> get_name_from_model_id(0x1000)
'Shelly Plus 1 UL'
```
**After:**
```python
>>> get_name_from_model_id(0x1000)
'Shelly Plus 1'
```
It is better to return the base model in this case (without `UL`)

Test output before this PR:
image

### DIFF
--- a/aioshelly/const.py
+++ b/aioshelly/const.py
@@ -544,7 +544,7 @@ DEVICES = {
         min_fw_date=GEN2_MIN_FIRMWARE_DATE,
         gen=GEN2,
         supported=True,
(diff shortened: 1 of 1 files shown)


bieniu (approved):
Looks good, thanks!

thecode:
Thanks!

PR #1027, opened 2026-01-02
Make http_request method in BlockDevice private

chemelli74:


thecode:
As I wrote on discord this is a breaking change for others and I don't see the benefit from this

thecode (changes requested):
I don't agree with this change as explained in the discussion

thecode:
Hi Simone,
I would like to explain again, hopefully you can accept my option.

As explained before I don't support this change.

I don't think we should block methods that are in device.py as I am using them outside of Home Assistant project and I know others are doing so too. `aioshelly` is not our private library and it has more users.
Even if it doesn't bother other users, it bothers me, I use the `call_rpc` as a side support for expanding the functionality of the Shelly irrigation controller which lakes many features I need, for this device using HTTP in parallel to WebSocket cause to much stress on the device until it crashes (already reported to Shelly but it doesn't matter there is no reason for me not to be possible to use `aioshelly` for other usages).

Regardless, I was never convinced why it is so important to make this method private, as we can decide what we accept or not for any feature usage of methods from `aioshelly` in Home Assistant, adding to that that as you noticed this also means we have to change the `set_state` inside the blocks which is just making the code less pretty.

**I think that adding tests is relevant and should be decoupled from this PR so that we can progress with expanding tests.**
I don't want you to spend time on this PR and eventually get frustrated that it is not merged or the other way around that it is merged and I will be.

We work on this library as small team with very few maintainers, thus we need to accept changes that support everyone in the team.

PR #1089, opened 2026-03-25
Bump orjson to 3.11.6 or newer

bieniu:
Security reason https://github.com/advisories/GHSA-hx9q-6w63-j58v

HA core repo uses orjson 3.11.7

https://github.com/home-assistant/core/blob/171b8dfa89e47550829dbe35c601b67dc006122a/requirements.txt#L37


bieniu:
Dependabot only bumps version if it is strictly pinned, e.g. `prek==0.3.6`

PR #1104, opened 2026-04-05
Improve rpc device tests

chemelli74:


bieniu (approved):

thecode (approved):

tests/rpc_device/conftest.py
```diff
@@ -90,7 +90,7 @@ async def _send_next_response(self) -> None:
 @pytest_asyncio.fixture
 async def rpc_websocket_responses() -> AsyncGenerator[ResponseMocker, None]:
     """Fixture for a WebSocket responses."""
-    return ResponseMocker()
+    yield ResponseMocker()
```
MartinHjelmare, line 93: There's no reason to yield here. We're not within a context manager. Same for the other changed lines in this PR. I'd revert and adjust the type annotation instead.

chemelli74: AFAIK `@pytest_asyncio.fixture` with `[AsyncGenerator]`need to use yield to properly set up and tear down resources.

MartinHjelmare: We only need to yield if we're inside a context manager as I wrote above. The type annotation was just incorrect before.

PR #1111, opened 2026-04-17
Handle mixed-case MAC address matching

benley:
If we end up with one side providing upper-case and the other a lower-case string form of the MAC address, that should still be considered a match.  Even better would be to parse and compare as a more specific datatype, but normalizing the strings is probably good enough for now.

### DIFF
--- a/tests/rpc_device/test_device.py
+++ b/tests/rpc_device/test_device.py
@@ -837,6 +837,34 @@ async def test_device_mac_address_mismatch(
 
 
 @pytest.mark.asyncio
+async def test_device_mac_address_mixed_case(
+    client_session: ClientSession,
+    ws_context: WsServer,
+    blu_gateway_device_info: dict[str, Any],
+    blu_gateway_config: dict[str, Any],
+    blu_gateway_status: dict[str, Any],
+    blu_gateway_remote_config: dict[str, Any],
+    blu_gateway_components: dict[str, Any],
+) -> None:
+    """Test RpcDevice initialize method when MAC differs in string case only."""
+    options = ConnectionOptions("10.10.10.10", device_mac="aabbccddeeff")
+
+    rpc_device = await RpcDevice.create(client_session, ws_context, options)
+    rpc_device.call_rpc_multiple = AsyncMock()
+
+    rpc_device.call_rpc_multiple.side_effect = [
+        [blu_gateway_device_info],
+        [blu_gateway_config, blu_gateway_status, blu_gateway_components],
+        [blu_gateway_remote_config],
+    ]
+
+    await rpc_device.initialize()
+
+    assert rpc_device.initialized is True
+    assert rpc_device.status is not None
+
+
+@pytest.mark.asyncio
 async def test_cover_update_status(
     rpc_device: RpcDevice,
     shelly2pmg3_status: dict[str, Any],
--- a/tests/test_common.py
+++ b/tests/test_common.py
@@ -115,6 +115,29 @@ async def test_get_info_mac_mismatch() -> None:
     await session.close()
 
 
+@pytest.mark.asyncio
+async def test_get_info_mac_mixed_case() -> None:
+    """Test get_info function when MAC differs in string case only."""
+    mock_response = await load_device_fixture("shellyplus2pm", "shelly.json")
+    ip_address = "10.10.10.10"
+
+    session = ClientSession()
+
+    with aioresponses() as session_mock:
+        session_mock.get(
+            URL.build(
+                scheme="http", host=ip_address, port=DEFAULT_HTTP_PORT, path="/shelly"
+            ),
+            payload=mock_response,
+        )
+
+        result = await get_info(session, ip_address, "aabbccddeeff")
+
+    await session.close()
+
+    assert result == mock_response
+
+
 @pytest.mark.parametrize(
     ("exc", "expected_exc"),
     [
--- a/aioshelly/common.py
+++ b/aioshelly/common.py
@@ -116,7 +116,7 @@ async def get_info(
         raise error from err
 
     mac = result["mac"]
-    if device_mac and device_mac != mac:
+    if device_mac and device_mac.lower() != mac.lower():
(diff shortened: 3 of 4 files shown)


benley:
This is an error I encountered in the wild, fwiw.

benley:
ah sorry about the failing test, I was too lazy to set up the test environment on my laptop so github was the first to run it. Trying to fix it now.

benley:
Should be fixed now, albeit a bit awkwardly.

bieniu:
> This is an error I encountered in the wild, fwiw.

Can you show us under what circumstances you experienced this error? Do you mean the real conditions related to Home Assistant integration when it happened?

benley:
> Can you show us under what circumstances you experienced this error? Do you mean the real conditions related to Home Assistant integration when it happened?

I just got a Wall Display XL, and home-assistant discovered it as an available device. When I clicked add, the integration failed with a MacAddressMismatchError in debug logs:

```
2026-04-17 16:29:39.639 DEBUG (MainThread) [aioshelly.rpc_device.device] host 192.168.7.241:80: error: MacAddressMismatchError('Input MAC: 02031AAC761B, Shelly MAC: 02031aac761b')
```

I don't know why it's getting a upper/lowercase difference, but those ought to be treated as the same MAC address behind the scenes.

thecode:
> I just got a Wall Display XL, and home-assistant discovered it as an available device. When I clicked add, the integration failed with a MacAddressMismatchError in debug logs:
> 
> ```
> 2026-04-17 16:29:39.639 DEBUG (MainThread) [aioshelly.rpc_device.device] host 192.168.7.241:80: error: MacAddressMismatchError('Input MAC: 02031AAC761B, Shelly MAC: 02031aac761b')
> ```

Can you check what is the output from `http://[device_ip]/shelly`? 

I think this PR is OK to procced but I would also like to understand if there is a difference (we already know Wall Display is not fully aligned with other Shelly devices API).

There is some work needed on the tests, I will comment directly on the places that needs to be fixed.

tests/rpc_device/test_device.py
```diff
@@ -836,6 +837,24 @@ async def test_device_mac_address_mismatch(
...
+    """Test RpcDevice initialize method when MAC differs in string case only."""
+    options = ConnectionOptions("10.10.10.10", device_mac="aabbccddeeff")
+
+    rpc_device = await RpcDevice.create(client_session, ws_context, options)
+    rpc_device.call_rpc_multiple = AsyncMock()
+
+    rpc_device.call_rpc_multiple.return_value = [blu_gateway_device_info]
+
+    with contextlib.suppress(IndexError):
+        await rpc_device.initialize()
```
thecode: It is not a good practice to suppress an error in a test since if something else will raise the same error we are missing it.

Test should also validate that the the `initialize` was successful, you can look at 
https://github.com/home-assistant-libs/aioshelly/blob/fe5ca13e2050b42008472f62f6c4c6ceaeeebc91/tests/rpc_device/test_device.py#L400-L401

as an example, it can also help as an example to let a device init correctly without raising `IndexError`

tests/test_common.py
```diff
@@ -115,6 +115,24 @@ async def test_get_info_mac_mismatch() -> None:
...
+    session = ClientSession()
+
+    with aioresponses() as session_mock:
+        session_mock.get(
+            URL.build(
+                scheme="http", host=ip_address, port=DEFAULT_HTTP_PORT, path="/shelly"
+            ),
+            payload=mock_response,
+        )
+        await get_info(session, ip_address, "aabbccddeeff")
```
thecode: To validate here you can use something similar to the original test:
https://github.com/home-assistant-libs/aioshelly/blob/fe5ca13e2050b42008472f62f6c4c6ceaeeebc91/tests/test_common.py#L71-L90

benley:
> > I just got a Wall Display XL, and home-assistant discovered it as an available device. When I clicked add, the integration failed with a MacAddressMismatchError in debug logs:
> > ```
> > 2026-04-17 16:29:39.639 DEBUG (MainThread) [aioshelly.rpc_device.device] host 192.168.7.241:80: error: MacAddressMismatchError('Input MAC: 02031AAC761B, Shelly MAC: 02031aac761b')
> > ```
> 
> Can you check what is the output from `http://[device_ip]/shelly`?
> 
> I think this PR is OK to procced but I would also like to understand if there is a difference (we already know Wall Display is not fully aligned with other Shelly devices API).
> 
> There is some work needed on the tests, I will comment directly on the places that needs to be fixed.

I can work on the tests again later this afternoon after some other work, thank you for the pointers.

This is the output from the device's /shelly endpoint:

```
{
  "id": "ShellyWallDisplay-02031aac761b",
  "mac": "02031aac761b",
  "model": "SAWD-3A1XE10EU2",
  "gen": 2,
  "fw_id": "20260416-104635/2.5.8-4ec844c8",
  "ver": "2.5.8",
  "app": "WallDisplayV2",
  "auth_en": false,
  "uptime": 52,
  "app_uptime": 27,
  "ram_size": 268435456,
  "ram_free": 104455368,
  "fs_size": 24480665600,
  "fs_free": 24070496256,
  "discoverable": true,
  "cfg_rev": 4,
  "schedule_rev": 0,
  "webhook_rev": 0,
  "platform": "vBlake.22f119c",
  "serial": "SA031AAC761BE",
  "batch_id": "3f811",
  "batch_date": 260113,
  "available_updates": {
    "beta": {
      "version": "2.6.0-beta1",
      "build_id": "20260415-073906/2.6.0-8b32eb27-beta1"
    }
  },
  "restart_required": false,
  "unixtime": 1776713999,
  "relay_in_thermostat": false,
  "relays_in_thermostat": [],
  "sensor_in_thermostat": false,
  "awaiting_auth_code": true,
  "ch": [
    "switch:0"
  ]
}

```

tests/test_common.py
benley: Fixed!

tests/rpc_device/test_device.py
benley: I believe this is fixed now, thanks again for the hints.

bieniu (approved):
Looks good. Thanks!

PR #1118, opened 2026-04-25
Fix RPC authentication for Shelly 2.0.0 firmware

thecode:
In firmware 2.0.0 Shelly improved device security - [2.0.0-beta1 Changelog](https://shelly-api-docs.shelly.cloud/gen2/changelog#security) mainly improving brute-force protection and preventing replay attacks.

This PR updates WsRPC class, when authentication is enabled calls are sent sequentially due to needed results from previous calls, without authentication calls are send in parallel (unchanged)

Fixes https://github.com/home-assistant/core/issues/169007

### DIFF
--- a/aioshelly/rpc_device/wsrpc.py
+++ b/aioshelly/rpc_device/wsrpc.py
@@ -9,7 +9,6 @@ import hashlib
 import logging
 import secrets
 import socket
-import time
 from asyncio import Task, tasks
 from collections.abc import Callable, Coroutine, Iterable
 from dataclasses import dataclass
@@ -113,15 +112,9 @@ class AuthData:
 
     def get_auth(self) -> dict[str, Any]:
         """Get auth for RPC calls with current nc value."""
-        # Generate cnonce based on nonce type
-        old_firmware = isinstance(self.nonce, int)
-        if old_firmware:
-            cnonce = int(time.time())
-        else:
-            random_bytes = secrets.token_bytes(16)
-            cnonce = base64.b64encode(random_bytes).decode("ascii")
+        random_bytes = secrets.token_bytes(16)
+        cnonce = base64.b64encode(random_bytes).decode("ascii")
 
-        # https://shelly-api-docs.shelly.cloud/gen2/Overview/CommonDeviceTraits/#authentication-over-websocket
         # Calculate response hash: SHA256(HA1:nonce:nc:cnonce:qop:HA2)
         hashed = hex_hash(f"{self.ha1}:{self.nonce}:{self.nc}:{cnonce}:auth:{HA2}")
 
@@ -136,8 +129,7 @@ class AuthData:
         }
 
         # Increment nc after each auth calculation
-        if not old_firmware:
-            self.nc += 1
+        self.nc += 1
 
         return frame
 
--- a/aioshelly/rpc_device/wsrpc.py
+++ b/aioshelly/rpc_device/wsrpc.py
@@ -4,7 +4,6 @@ from __future__ import annotations
 
 import asyncio
 import base64
-import contextlib
 import hashlib
 import logging
 import secrets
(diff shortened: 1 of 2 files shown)


aioshelly/rpc_device/wsrpc.py
```diff
@@ -424,60 +441,91 @@ def _raise_for_unrecoverable_errors(
...
+                        method,
+                        params,
+                        timeout,
+                        allow_auth_retry=False,
+                    )
+                call.result = response["result"]
+        except TimeoutError as exc:
+            with contextlib.suppress(asyncio.CancelledError):
+                call.resolve.cancel()
+                await call.resolve
```
bdraco: needs to avoid swallowing cancellation
```
if (current_task := asyncio.current_task() and current_task.cancelling()):
  raise
```

aioshelly/rpc_device/wsrpc.py
```diff
@@ -174,14 +189,18 @@ def __repr__(self) -> str:
     @property
     def request_frame(self) -> dict[str, Any]:
```
bdraco: this property now has side effects because it increments nc. That might lead to accidental bugs

aioshelly/rpc_device/wsrpc.py
```diff
@@ -424,60 +441,91 @@ def _raise_for_unrecoverable_errors(
...
-                    auth["nonce"], auth.get("nc", 1)
+        if self._session.auth_data is not None:
+            # sequential: each call uses fresh nc from prior challenge,
+            # first call seeds the nonce via 401 challenge
+            results = []
+            for method, params in calls:
+                results.append(
+                    await self._rpc_call_with_auth_retry(method, params, timeout)
                 )
-            to_retry.append(call)
```
bdraco, line 464: small behavior change here. the timeout is now per call instead of for all calls.

If we want to keep it per all calls we could keep track of deadline and subtract between calls

ie

```python
  deadline = self._loop.time() + timeout                                                                                                                                                                                            
  for method, params in calls:                                      
      remaining = deadline - self._loop.time()
      results.append(await self._rpc_call_with_auth_retry(method, params, remaining))
```

aioshelly/rpc_device/wsrpc.py
```diff
@@ -424,60 +441,91 @@ def _raise_for_unrecoverable_errors(
...
-            if (result := call.result) is UNDEFINED:
-                resp = call.resolve.result()
-                self._raise_for_unrecoverable_errors(resp, allow_auth_retry=False)
-            else:
-                successful.append(result)
+            return results
 
-        return successful
+        results = await self._rpc_calls(calls, timeout)
+        return [call.result for call in results if call.result is not UNDEFINED]
```
bdraco:
```suggestion
        return [call.result for call in results]
```

looks unreachable now

aioshelly/rpc_device/wsrpc.py
```diff
@@ -95,44 +96,58 @@ class AuthData:
...
 
-    def get_auth(self, nonce: int | None = None, n_c: int = 1) -> dict[str, Any]:
-        """Get auth for RPC calls."""
-        cnonce = int(time.time())
-        if nonce is None:
-            nonce = cnonce - 1800
+    def update_challenge(self, auth_challenge: dict[str, Any]) -> None:
+        """Update authentication challenge data from server."""
+        self.nonce = auth_challenge["nonce"]
+        self.algorithm = auth_challenge["algorithm"]
```
bdraco: probably should raise if algo is unexpected as we are going to send sha256 reguardless

aioshelly/rpc_device/wsrpc.py
bdraco: note that we have this problem in other places

bieniu (approved):
Looks good 👍 Runs all day on my prod without any problems.

PR #1123, opened 2026-04-28
Update supports_scripts to check against Shelly.ListMethods

thecode:
I tested this with multiple devices:
- Standard Gen2/3/4 relay devices that supports scripts with different firmware versions from 1.3 to 2.0
- Wall Display with firmware 2.3.3 (does not supports scripts)
- ST1820 with firmware 1.4.99-xt-prod1 (does not supports scripts)
- Wall Display with firmware 2.6.0 (supports scripts)

This fixes https://github.com/home-assistant/core/issues/168292, I did notice other problems when testing this device, but they will not prevent the device from working and needed to be fixed in Home Assistant side.


bieniu (approved):
Looks good 👍 Thanks

chemelli74:
Tested wit ha few devices and works fine. Nice one !

chemelli74 (approved):

PR #1125, opened 2026-04-29
Ensure RPC auth challenge nc is an integer

thecode:
Shelly Wall Display uses string for `nc` counter during auth challenge, casting to int works on all devices, tested with:
- Wall Display firmware 2.3.3
- Wall Display firmware 2.6.0-beta
- Shelly Pro 1 firmware 1.4.4
- Shelly 1PM Gen3 2.0.0-beta1

### DIFF
--- a/aioshelly/rpc_device/wsrpc.py
+++ b/aioshelly/rpc_device/wsrpc.py
@@ -114,7 +114,7 @@ class AuthData:
             raise InvalidAuthError(
                 f"Unsupported auth algorithm: {auth_challenge['algorithm']}"
             )
-        self.nc = auth_challenge.get("nc", 1)
+        self.nc = int(auth_challenge.get("nc", 1))
 
     def get_auth(self) -> dict[str, Any]:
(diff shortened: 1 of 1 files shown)


chemelli74 (approved):

aioshelly/rpc_device/wsrpc.py
```diff
@@ -114,7 +114,7 @@ def update_challenge(self, auth_challenge: dict[str, Any]) -> None:
             raise InvalidAuthError(
                 f"Unsupported auth algorithm: {auth_challenge['algorithm']}"
             )
-        self.nc = auth_challenge.get("nc", 1)
+        self.nc = int(auth_challenge.get("nc", 1))
```
bieniu, line 119: Could we add a comment here explaining why `int()` is needed?

bieniu: Thanks!

bieniu (approved):

PR #1126, opened 2026-04-29
Use RPC authentication only if enabled on the device

thecode:
When testing authentication I enabled it for few devices on my prod and later disabled it on all of them.
I noticed we still do the authentication process and send call sequentially instead of parallel calls since Home Assistant doesn't know it was disabled on the device.

Added a check to use authentication only if enabled on the device (`auth_en` flag from `Shelly.GetDeviceInfo` endpoint)

We don't need to change anything on Home Assistant side for this, if the user disable the authentication we ignore the provided credentials so the user doesn't need to do anything.

Tested on:
- Wall Display firmware 2.3.3
- Wall Display firmware 2.6.0-beta
- Shelly Pro 1 firmware 1.4.4
- Shelly 1PM Gen3 2.0.0-beta1

### DIFF
--- a/aioshelly/rpc_device/device.py
+++ b/aioshelly/rpc_device/device.py
@@ -674,7 +674,8 @@ class RpcDevice:
         self._shelly = await self.call_rpc("Shelly.GetDeviceInfo")
         # Auth only supported on WebSocket transport
         if (
-            self.options.username
+            self.requires_auth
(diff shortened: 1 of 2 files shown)


chemelli74 (approved):
Nice catch !

bieniu (approved):
Great catch!

PR #1135, opened 2026-05-10
Add ability to update Uart/Lora Add-on

chemelli74:
_Current official UART Add-On version - "2.1.1" supports: EU868, US915, BR915 RF plans for [Shelly LoRa](https://shelly-api-docs.shelly.cloud/gen2/Addons/ShellyLoRaAddon)_

https://shelly-api-docs.shelly.cloud/gen2/Addons/ShellyUartAddon#addonupdate-example

### DIFF
--- a/aioshelly/rpc_device/device.py
+++ b/aioshelly/rpc_device/device.py
@@ -326,6 +326,11 @@ class RpcDevice:
         params = {"stage": "beta"} if beta else {"stage": "stable"}
         await self.call_rpc("Shelly.Update", params)
 
+    async def trigger_add_on_ota_update(self, timeout: int = 300) -> None:
+        """Trigger an add-on ota update."""
+        params = {"timeout": timeout}
+        await self.call_rpc("AddOn.Update", params)
+
     async def trigger_reboot(self, delay_ms: int = 1000) -> None:
         """Trigger a device reboot."""
         await self.call_rpc("Shelly.Reboot", {"delay_ms": delay_ms})
--- a/tests/rpc_device/test_device.py
+++ b/tests/rpc_device/test_device.py
@@ -1045,6 +1045,16 @@ async def test_trigger_ota_update(rpc_device: RpcDevice) -> None:
 
 
 @pytest.mark.asyncio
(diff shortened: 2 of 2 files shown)


aioshelly/rpc_device/device.py
```diff
@@ -321,10 +321,17 @@ def subscribe_updates(self, update_listener: Callable) -> None:
...
 
-    async def trigger_ota_update(self, beta: bool = False) -> None:
+    async def trigger_ota_update(
+        self, beta: bool = False, add_on: bool = False
+    ) -> None:
         """Trigger an ota update."""
-        params = {"stage": "beta"} if beta else {"stage": "stable"}
-        await self.call_rpc("Shelly.Update", params)
+        if add_on:
+            params = {"timeout": 600}
```
bieniu: Why do we need this timeout? Docs says that the default timeout is 300 ms.

aioshelly/rpc_device/device.py
```diff
@@ -321,10 +321,17 @@ def subscribe_updates(self, update_listener: Callable) -> None:
         """Subscribe to device status updates."""
         self._update_listener = update_listener
 
-    async def trigger_ota_update(self, beta: bool = False) -> None:
+    async def trigger_ota_update(
```
bieniu: I wonder if it wouldn't be better to implement a separate method for updating the addon 🤔

aioshelly/rpc_device/device.py
chemelli74: I just used the example they provide ;-)

aioshelly/rpc_device/device.py
chemelli74: We can, but in the end is the same topic and makes HA implementation simple

thecode: It doesn't make sense to me to use the same python method when the device uses a different method, at first I thought it is another parameter of the update but if it is another method I think it would be better to add another python method (e.g. `trigger_addon_ota`)

chemelli74: Done

thecode (approved):

chemelli74:
Thx !

PR #1136, opened 2026-05-10
Introduce `add_on_installed()` and `add_on_info()` methods

chemelli74:
- Reply when LoRa/UART add-on is not installed:
```json
{
  "code": 404,
  "message": "No handler for AddOn.GetInfo"
}
```
- Extract of config when LoRa/UART Add-On is installed:
```json
  "sys": {
    "device": {
      "name": "LoRa - transmitter",
      "mac": "XXXXXXXXXXXX",
      "fw_id": "20260311-095901/1.7.5-g9979d16",
      "discoverable": true,
      "eco_mode": false,
      "addon_type": "LoRa"
    },
```
- Extract of config when Sensor Add-On is installed:
```json
  "sys": {
    "device": {
      "name": "Irrigazione",
      "mac": "YYYYYYYYYYYY",
      "fw_id": "20260423-102620/2.0.0-beta1-g8c7700a",
      "discoverable": true,
      "eco_mode": false,
      "addon_type": "sensor",
      "tls_check_cert_validity_time": true,
      "enhanced_security": false
    },
```

Doc example:
- <https://shelly-api-docs.shelly.cloud/gen2/Addons/ShellyUartAddon#addongetinfo-example>


aioshelly/rpc_device/device.py
```diff
@@ -1009,6 +1012,34 @@ def zigbee_firmware(self) -> bool:
 
         return "zigbee" in self._config
 
+    @property
+    def add_on_installed(self) -> bool:
+        """Return True if add-on is installed."""
+        if self.gen not in RPC_GENERATIONS:
+            return False
```
thecode: What is the point of this check? how can `self.gen` not be part of RPC generations?

chemelli74: if gen is GEN1 for example ?

But we are calling the method only for RPC devices ;-)
Will remove it

aioshelly/rpc_device/device.py
```diff
@@ -1014,6 +1016,26 @@ def zigbee_firmware(self) -> bool:
 
         return "zigbee" in self._config
 
+    @property
+    def add_on_installed(self) -> bool:
```
bieniu: Why add new property based on config if config is available in the integration code?

aioshelly/rpc_device/device.py
```diff
@@ -1014,6 +1016,26 @@ def zigbee_firmware(self) -> bool:
...
+        """Return True if add-on is installed."""
+        return self.config["sys"]["device"].get("addon_type") is not None
+
+    async def add_on_info(self) -> dict[str, Any]:
```
bieniu, line 1021: The purpose of this method is unclear for me. It is intended to return information about the LoRa addon, but also to return the addon type (this information is available via config). Two different functionalities are combined here.

aioshelly/rpc_device/device.py
```diff
@@ -1014,6 +1016,26 @@ def zigbee_firmware(self) -> bool:
...
+        if not self.add_on_installed:
+            return {}
+
+        # Shelly Sensor Add-On
+        if self.config["sys"]["device"]["addon_type"] == "sensor":
+            return {"type": "Sensor"}
+
+        # Shelly Uart / LoRa Add-On
+        info = await self.call_rpc("AddOn.GetInfo")
+        if info.get("code") == RPC_CALL_ERR_NO_HANDLER:
```
thecode: How does this work? when we used the `RPC_CALL_ERR_NO_HANDLER` in scripts detection method it actually raise an error which we catch

aioshelly/rpc_device/device.py
chemelli74: Like for zigbee  just a couple of lines above ?

aioshelly/rpc_device/device.py
chemelli74: It's intened to return information of whatever add-on is installed, now and in future.
Not only for LoRa one.

I don't get the point. Can you please clarify it ?

aioshelly/rpc_device/device.py
chemelli74: this is not raisyng and error anymore, just returning a json with that info.
You can test with a gen2+ device with no add-on installed.

thecode: It does raise, I added `AddOn.GetInfo` to the calls we sent to the device:
```
DEBUG:aioshelly.rpc_device.device:host 192.168.1.82:80: error: DeviceConnectionError(RpcCallError(404, 'No handler for AddOn.GetInfo'))
DEBUG:aioshelly.rpc_device.wsrpc:Websocket client connection from 192.168.1.82:80 closed
Error connecting to 192.168.1.82:80, error: DeviceConnectionError(RpcCallError(404, 'No handler for AddOn.GetInfo'))
```

chemelli74: Interesting, not here 🤔 

I'll update the code with the try/catch just to be safe

chemelli74: I added a call to `add_on_info()` in example.py and tested with 3 devices: 
- gen2 with sensor
- gen4 with lora
- gen3 without 

and all works fine here

aioshelly/rpc_device/device.py
bieniu: OK, good point. But we use this property in the integration code. How do you want to use this property in the integration code?

aioshelly/rpc_device/device.py
bieniu: My point is I don't understand what value bring `{"type": "Sensor"}` to us.

- Why do we need this?
- How do we want to use this?
- Why is the value `Sensor` and not `sensor`?

chemelli74: is for exposing the info of which tpye of sensors is installed on the device.
What will be the exact usage on HomeAssistant will still need to be undestood.

aioshelly/rpc_device/device.py
chemelli74: Like first code line of `add_on_info()` ?

aioshelly/rpc_device/device.py
chemelli74: Here my code in example,py that will be very close to the way HomeAssistant will use this method:

image

and I cannot see any exception running it.
Could you please test this way and report back ?
Thank you

aioshelly/rpc_device/device.py
bieniu: And this one line requires us to add new property? Wouldn't it be better to check the config on this line?

aioshelly/rpc_device/device.py
bieniu: IMO once we understand how we want to use it then we should implement it appropriately, without hardcoding `{"type": "Sensor"}`.

aioshelly/rpc_device/device.py
chemelli74: again, why we wanted it for zigbee then ? We should remove it as well isn't it ?

aioshelly/rpc_device/device.py
chemelli74: @thecode any kind feedback on your test ?

thecode: Nothing has changed from https://github.com/home-assistant-libs/aioshelly/pull/1136#discussion_r3224910184 your example above never reach line 1035, if it will reach it, it will raise.
Can you supply the exact scenario which made you add `if info.get("code") == RPC_CALL_ERR_NO_HANDLER:` so I can test it?

chemelli74: As I wrote, try using` device.add_on_info()` on a gen2+ device with no add-on installed.
I get no exception and the json reply you see on the PR description.

thecode: The JSON reply in the PR description should raise, I am assuming you took it from http call and not from our `WsRPC` class.

Can you run the example in with debug (`--deb`) flag and show the output where line 1035 is evaluated and not raising?

chemelli74: Will do, but if it raises than the call I already do device.add_on_info() should trace isn't it ?

aioshelly/rpc_device/device.py
bieniu: Yes, you can remove the `zigbee_enabled` property since it is not used in the integration code but this will be a breaking change. IMO it is not worth creating a breaking change just to remove one property.

I have no problem adding the `add_on_installed` property (or any other) if it will be used in the integration code and will simplify that code. There's currently no draft PR that requires this property. And if there is I'm not aware of it. So at this point, adding a property that does nothing other than reading the config doesn't make sense from my point of view.

aioshelly/rpc_device/device.py
bluetoothbot: Yes — it raises, and tracing the code path shows why your example doesn't seem to.

`device.add_on_info()` → `call_rpc("AddOn.GetInfo")` → `call_rpc_multiple()` → `WsRPC.calls()` → `_rpc_call_with_auth_retry()`. When the device replies `{"error": {"code": 404, "message": "No handler for AddOn.GetInfo"}}` there is no `result` key, so `_raise_for_unrecoverable_errors()` runs (`wsrpc.py:440-452`). Because `code` (404) is not 401/UNAUTHORIZED, it raises `RpcCallError(404, "No handler for AddOn.GetInfo")`. `call_rpc_multiple()` catches and re-raises it (`device.py:901-903`).

So two things follow:
1. The exception propagates out of `add_on_info()` before line 1035 — the method does not return `{}` for a non-installed UART/LoRa add-on, it raises.
2. `if info.get("code") == RPC_CALL_ERR_NO_HANDLER` can never be true: on success `call_rpc` returns only the `result` payload (`wsrpc.py:497`, `:539`), which has no top-level `code` field.

The reason it looks fine when you test manually is almost certainly that you're observing the raw HTTP/JSON body, not the value the library hands back through `WsRPC`. The HTTP endpoint returns the error JSON verbatim; `WsRPC` turns it into an exception.

The established fix is the same pattern already used in `aioshelly/ble/__init__.py:61-65`:
```python
try:
    return await self.call_rpc("AddOn.GetInfo")
except RpcCallError as err:
    if err.code == RPC_CALL_ERR_NO_HANDLER:
        return {}
    raise
```
To reproduce the raise in a test, set `call_rpc_multiple.side_effect = RpcCallError(404, "No handler for AddOn.GetInfo")` rather than a `return_value`.

aioshelly/rpc_device/device.py
bluetoothbot: On the `Sensor` vs `sensor` point specifically: the device reports the field lowercase — the PR description shows `"addon_type": "sensor"`, and the implementation itself compares against the lowercase literal `"sensor"` (line 1030). Returning `{"type": "Sensor"}` (capitalized) is therefore inconsistent both with the device's own value and with the rest of the method, which returns the device's raw `AddOn.GetInfo` payload verbatim for UART/LoRa. So today `add_on_info()` hands back two different shapes depending on add-on type.

If the sensor case is meant to short-circuit the RPC, it'd be more consistent (and avoid the hardcode you flagged) to derive the value from config, e.g. `{"type": config["sys"]["device"]["addon_type"]}`, or to define a single normalized shape that both branches return. Agree with your broader point that the contract should be settled once the integration use case is understood, rather than hardcoding `{"type": "Sensor"}`.

aioshelly/rpc_device/device.py
chemelli74: Removed, even if I think we should be consistent on the approach.

aioshelly/rpc_device/device.py
chemelli74: Type will be used on HA side to show the user what is installed so

```python
{"type": XXX }
```

will be consistent and easy to retrieve.

I removed the case normalization anyway.

aioshelly/rpc_device/device.py
chemelli74: There is no exception because method calls `AddOn.GetInfo` only if an add-on is installed.

So the Error check is a dead end ;-)

Removed

bluetoothbot: For what it's worth, the code confirms @thecode's analysis. The chain is `add_on_info()` → `call_rpc("AddOn.GetInfo")` → `call_rpc_multiple(...)`, and `call_rpc_multiple` explicitly re-raises `RpcCallError` for device errors (device.py:905-907):

```python
except (InvalidAuthError, RpcCallError) as err:
    self._last_error = err
    raise
```

So a device returning `404 No handler for AddOn.GetInfo` over `WsRPC` surfaces as a **raised** `RpcCallError(404, ...)`, never as a returned `{"code": 404, ...}` dict. The JSON in the PR description was almost certainly captured from a raw HTTP call, not from the `WsRPC` path.

The reason the `example.py` run never raised is that line 1034 is only reached when `addon_type` is a truthy non-`sensor` value — i.e. an add-on is installed and its handler exists. The gen3-without-addon case returns early at `return {}`, and the sensor case returns before any RPC. So the removed `if info.get("code") == RPC_CALL_ERR_NO_HANDLER` really was dead code, as you concluded.

Two leftovers to clean up, though: (1) `RPC_CALL_ERR_NO_HANDLER` (line 59) is now defined but unused; (2) the `uart_error` test case feeds `{"code": 404, ...}` as a *return* value from `call_rpc_multiple`, which can't happen on a real device. If you're confident the 'installed ⇒ handler exists' invariant holds across all firmware, drop both. If you want defensive handling of a missing handler, catch `RpcCallError`, check `err.code == RPC_CALL_ERR_NO_HANDLER` (using the constant), and make the test use `side_effect=RpcCallError(404, ...)`.

aioshelly/rpc_device/device.py
bluetoothbot: Agree the return contract is the real crux here. As implemented, `add_on_info()` hands back three different shapes: `{}` when nothing is installed, a synthesized `{"type": "sensor"}` for the sensor add-on, and the raw `AddOn.GetInfo` payload (with its own keys) for UART/LoRa. A caller can't branch on a single stable field across add-on types.

A path forward that avoids the hardcoded one-off while still being useful: always return a normalized dict keyed on `type`. The sensor branch returns `{"type": add_on_type}`; the UART/LoRa branch returns `{"type": add_on_type, **rpc_info}` (merging the config-derived type with the `AddOn.GetInfo` fields). Then `result.get("type")` is meaningful for every add-on, the value is no longer hardcoded, and HA can branch on one key regardless of which add-on is present. That said, it would be best to nail down the intended HA consumption before locking this in as a public method — without a concrete consumer the schema is guesswork.

chemelli74: Code reviewed

tests/rpc_device/test_device.py
```diff
@@ -2122,3 +2122,56 @@ async def test_media_list_radio_stations(
 
     assert isinstance(result, list)
     assert len(result) == 2
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize(
+    ("addon_type", "methods", "rpc_response", "expected", "call_count"),
+    [
+        pytest.param(None, [], None, {"type": None}, 1, id="not_installed"),
```
bieniu, line 2147: We haven't used `pytest.param` before, I don't think it provides any benefits here.

tests/rpc_device/test_device.py
```diff
@@ -928,9 +928,9 @@ async def test_connect_websocket_reraises_mac_mismatch(rpc_device: RpcDevice) ->
     rpc_device._init_calls = AsyncMock(side_effect=error)
 
     with (
-        patch.object(rpc_device._rpc, "connect", wraps=rpc_device._rpc.connect),
+        patch.object(rpc_device._rpc, "connect", new_callable=AsyncMock),
```
bieniu: Why do we want to change this?

tests/test_common.py
```diff
@@ -100,14 +100,15 @@ async def test_process_ip_or_options() -> None:
...
+async def test_get_info(aiohttp_json_response_ctx: Any) -> None:
     """Test get_info function."""
     mock_response = await load_device_fixture("shellyplus2pm", "shelly.json")
     ip_address = "10.10.10.10"
 
     session = ClientSession()
+    request_ctx = aiohttp_json_response_ctx(mock_response)
 
-    with mock_shelly_get(payload=mock_response):
+    with patch.object(session, "get", return_value=request_ctx):
```
bieniu: Why do we want to change this?

tests/rpc_device/test_device.py
chemelli74: Nick requested to add "id"

tests/rpc_device/test_device.py
chemelli74: Unwanted change, reverted

tests/test_common.py
chemelli74: TO align with other tests. Anyway, reverted

chemelli74: Full reverted

bieniu (approved):
Looks good, thanks!

PR #1137, opened 2026-05-10
Fix directory typo in build script

chemelli74:
Repository is based on `.venv` not `venv`.

### DIFF
--- a/script/environment.sh
+++ b/script/environment.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 pip3 --disable-pip-version-check --no-cache-dir install uv
-uv venv venv --seed --clear
-source venv/bin/activate
+uv venv .venv --seed --clear
+source .venv/bin/activate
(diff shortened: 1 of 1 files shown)


script/environment.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 pip3 --disable-pip-version-check --no-cache-dir install uv
-uv venv venv --seed --clear
```
bieniu, line 4: This is not a typo

https://github.com/home-assistant-libs/aioshelly/blob/0c122cb1569c5ddd274c83352e4082594a19813b/.pre-commit-config.yaml#L60

https://github.com/home-assistant-libs/aioshelly/blob/0c122cb1569c5ddd274c83352e4082594a19813b/.devcontainer/devcontainer.json#L16

chemelli74: I ended up with both .venv and venv somehow.

- deactivate
- rm -rf .venv
- rm -rf venv
- rebuild container

and now everything looks good.

PR #1147, opened 2026-05-22
Implement on-demand active scan window on ShellyBLEScanner

bdraco:
Pairs with Bluetooth-Devices/habluetooth#448 and home-assistant/core#172008. habluetooth gained a per-device active-scan scheduler that flips AUTO-mode scanners into active for a short window on demand; for that to work on Shelly proxies the ShellyBLEScanner needs to know how to flip its BLE script's scan mode.

## Behavior

`ShellyBLEScanner.async_request_active_window` flips the device's BLE scanner to active via `Script.Eval`, sleeps for the requested duration, then flips it back to passive. Callers wire the `RpcDevice` via `set_active_window_provider(device)`; without that the override returns `False` so it matches the no-op default on `BaseHaScanner` and existing setups keep working unchanged. A per-scanner `asyncio.Lock` serializes overlapping requests so the device only ever sees one mode transition pair at a time, `ShellyError` during entry or restore is logged and swallowed, and the scanner caches the BLE script id after the first window so every subsequent flip is just one Script.Eval round-trip.

## Why Script.Eval instead of re-flashing

The first cut re-ran `async_start_scanner` per window, which re-templated the body with the new `active=true|false` literal, detected the body diff, and issued `Script.PutCode` (a flash write) twice per window. For a power feature with potentially frequent windows that's noticeable flash wear. The BLE script now exposes a `setActive(v)` JS function, and the new `async_set_active_mode` helper drives it through `Script.Eval`. Window flips touch zero flash and the steady state is exactly two `Script.Eval` calls per window.

Two Shelly Script API quirks shape the JS side:
1. `BLE.Scanner.Stop()` is asynchronous, so a `Start()` in the same JS frame races with the still-in-progress stop and silently fails. `setActive` defers the `Start` via `Timer.set(250, ...)`.
2. `BLE.Scanner.Subscribe()` does not survive a Stop/Start cycle, so it is re-attached inside `applyPendingActive` after every `Start`.

Script header bumped to 2.1; existing devices migrate once on the next `async_start_scanner` (one putcode), then run flash-free for every window thereafter.

## Hardware verification

Tested on Shelly Plug US (SNPL-00116US, fw 1.7.5):
- One-time 2.0 to 2.1 migration: exactly one `Script.PutCode`
- Per window: exactly 2 `Script.Eval`, zero `PutCode`/`Start`/`Stop`
- Script id resolved exactly once across multiple windows on the same scanner
- Script body byte-identical across consecutive windows
- Active-mode actually flips: a background subscriber capturing `ble.scan_result` events to a JSONL file shows `scan_rsp` populated for ~40-50% of advertisements during active windows and ~5-10% during passive (residue is queue spill from samples emitted just after the flip)

## Status

Draft because it depends on habluetooth 6.3.0, which is not released yet.

### DIFF
--- a/aioshelly/ble/__init__.py
+++ b/aioshelly/ble/__init__.py
@@ -75,6 +75,27 @@ async def async_start_scanner(
     await device.script_start(ble_script_id)
 
 
+async def async_set_active_mode(device: RpcDevice, active: bool) -> None:
+    """Flip the BLE scanner's active mode via Script.Eval.
+
+    The script body exposes a ``setActive(v)`` function (see BLE_CODE
+    in const.py) that stops and restarts BLE.Scanner with the new
+    active flag. Driving the toggle via Script.Eval keeps the script
+    body byte-identical across windows, so the device's flash is not
+    rewritten on every transition.
+
+    Assumes the script is already installed and running via
+    ``async_start_scanner``. Raises RpcCallError if the script is not
+    installed on the device.
+    """
+    script_name_to_id = await _async_get_scripts_by_name(device)
+    if (ble_script_id := script_name_to_id.get(BLE_SCRIPT_NAME)) is None:
+        raise RpcCallError(0, f"{BLE_SCRIPT_NAME} script not installed")
+    await device.script_eval(
+        ble_script_id, f"setActive({'true' if active else 'false'})"
+    )
+
+
 def create_scanner(
     source: str,
     name: str,
--- a/aioshelly/ble/__init__.py
+++ b/aioshelly/ble/__init__.py
@@ -75,7 +75,21 @@ async def async_start_scanner(
     await device.script_start(ble_script_id)
 
 
-async def async_set_active_mode(device: RpcDevice, active: bool) -> None:
+async def async_get_ble_script_id(device: RpcDevice) -> int | None:
+    """Return the installed BLE integration script id, or None if absent.
+
+    Callers that toggle active mode repeatedly (e.g. the active-window
+    scheduler) should resolve the id once and cache it for the
+    lifetime of the script, since the id is stable until the script
+    is recreated.
+    """
+    script_name_to_id = await _async_get_scripts_by_name(device)
+    return script_name_to_id.get(BLE_SCRIPT_NAME)
+
+
+async def async_set_active_mode(
+    device: RpcDevice, script_id: int, active: bool
+) -> None:
     """Flip the BLE scanner's active mode via Script.Eval.
 
     The script body exposes a ``setActive(v)`` function (see BLE_CODE
@@ -84,16 +98,11 @@ async def async_set_active_mode(device: RpcDevice, active: bool) -> None:
     body byte-identical across windows, so the device's flash is not
     rewritten on every transition.
 
-    Assumes the script is already installed and running via
-    ``async_start_scanner``. Raises RpcCallError if the script is not
-    installed on the device.
+    ``script_id`` must be the id returned by ``async_get_ble_script_id``
+    or otherwise known to the caller. This keeps each flip a single
+    Script.Eval round-trip with no Script.List overhead.
     """
-    script_name_to_id = await _async_get_scripts_by_name(device)
-    if (ble_script_id := script_name_to_id.get(BLE_SCRIPT_NAME)) is None:
-        raise RpcCallError(0, f"{BLE_SCRIPT_NAME} script not installed")
-    await device.script_eval(
-        ble_script_id, f"setActive({'true' if active else 'false'})"
-    )
+    await device.script_eval(script_id, f"setActive({'true' if active else 'false'})")
 
 
 def create_scanner(
--- a/aioshelly/ble/backend/scanner.py
+++ b/aioshelly/ble/backend/scanner.py
@@ -56,7 +56,7 @@ class ShellyBLEScanner(BaseHaRemoteScanner):
         if device is None or event_type is None or data_version is None:
             return False
         # Import here to avoid a circular import with aioshelly.ble.
-        from aioshelly import ble as _ble
+        from aioshelly import ble as _ble  # noqa: PLC0415
 
         async with self._active_window_lock:
             try:
--- a/aioshelly/ble/backend/scanner.py
+++ b/aioshelly/ble/backend/scanner.py
@@ -49,13 +49,16 @@ class ShellyBLEScanner(BaseHaRemoteScanner):
 
         Called by habluetooth's auto-mode scheduler. The Shelly BLE
         script is reprovisioned in active mode, then reverted to passive
-        once the window ends. Concurrent requests are serialized so the
-        device only sees one mode transition pair at a time.
+        once the window ends. Only one window may be open at a time; a
+        request that arrives while another window is in flight returns
+        ``False`` immediately so the caller can decide whether to retry.
         """
         device = self._active_window_device
         event_type = self._active_window_event_type
         data_version = self._active_window_data_version
         if device is None or event_type is None or data_version is None:
+            return False
+        if self._active_window_lock.locked():
             return False
 
         async with self._active_window_lock:
@@ -74,12 +77,17 @@ class ShellyBLEScanner(BaseHaRemoteScanner):
             try:
                 await asyncio.sleep(duration)
             finally:
+                # Shield the restore so a cancellation mid-window still
+                # reverts the device to passive instead of leaving it
+                # burning battery in active mode.
                 try:
-                    await _ble.async_start_scanner(
-                        device,
-                        active=False,
-                        event_type=event_type,
-                        data_version=data_version,
+                    await asyncio.shield(
+                        _ble.async_start_scanner(
+                            device,
+                            active=False,
+                            event_type=event_type,
+                            data_version=data_version,
+                        )
                     )
                 except RpcCallError as err:
                     LOGGER.debug(
--- a/aioshelly/ble/backend/scanner.py
+++ b/aioshelly/ble/backend/scanner.py
@@ -11,7 +11,7 @@ from habluetooth import BaseHaRemoteScanner
 
 from aioshelly import ble as _ble
 
-from ...exceptions import RpcCallError
+from ...exceptions import ShellyError
 from ..const import BLE_SCAN_RESULT_EVENT
 from ..parser import parse_ble_scan_result_event
 
@@ -69,7 +69,7 @@ class ShellyBLEScanner(BaseHaRemoteScanner):
                     event_type=event_type,
                     data_version=data_version,
                 )
-            except RpcCallError as err:
+            except ShellyError as err:
                 LOGGER.debug(
                     "%s: failed to enter active scan window: %s", self.name, err
                 )
@@ -89,7 +89,7 @@ class ShellyBLEScanner(BaseHaRemoteScanner):
                             data_version=data_version,
                         )
                     )
-                except RpcCallError as err:
+                except ShellyError as err:
                     # Restore failures leave the device stuck in active
                     # mode (drawing power, contradicting requested_mode
                     # in habluetooth) until the next successful window;
--- a/aioshelly/ble/backend/scanner.py
+++ b/aioshelly/ble/backend/scanner.py
@@ -90,7 +90,12 @@ class ShellyBLEScanner(BaseHaRemoteScanner):
                         )
                     )
                 except RpcCallError as err:
-                    LOGGER.debug(
+                    # Restore failures leave the device stuck in active
+                    # mode (drawing power, contradicting requested_mode
+                    # in habluetooth) until the next successful window;
+                    # surface that at warning so operators can see it
+                    # without flipping the package to debug.
+                    LOGGER.warning(
                         "%s: failed to restore scan mode after active window: %s",
                         self.name,
                         err,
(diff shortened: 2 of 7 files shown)


aioshelly/ble/backend/scanner.py
```diff
@@ -2,21 +2,93 @@
...
+
+    async def async_request_active_window(self, duration: float) -> bool:
+        """Flip the Shelly's BLE script to active for ``duration`` seconds.
+
+        Called by habluetooth's auto-mode scheduler. The Shelly BLE
+        script is reprovisioned in active mode, then reverted to passive
+        once the window ends. Concurrent requests are serialized so the
+        device only sees one mode transition pair at a time.
+        """
+        device = self._active_window_device
```
bdraco, line 52: probably need a reentrency guard just in case but habluetooth already promises that .. but technically its a public api

aioshelly/ble/backend/scanner.py
```diff
@@ -2,21 +2,93 @@
...
+        event_type = self._active_window_event_type
+        data_version = self._active_window_data_version
+        if device is None or event_type is None or data_version is None:
+            return False
+        # Import here to avoid a circular import with aioshelly.ble.
+        from aioshelly import ble as _ble
```
bdraco: ideally we avoid late import but might need to happen in a followup as the refactor could get large

aioshelly/ble/backend/scanner.py
bluetoothbot: The `asyncio.Lock` already gives you serialization across concurrent tasks, which is the only thing a public-API caller can realistically do to break this — you can't accidentally reenter from the same task without explicitly spawning another one. If you want belt-and-suspenders against the same-task reentry case (which would currently deadlock on the lock acquire rather than misbehave), the cheapest option is an `asyncio.Lock` paired with a fail-fast check:

```python
if self._active_window_lock.locked():
    return False  # or: extend the existing window
async with self._active_window_lock:
    ...
```

That turns the contract from "queued behind the in-flight window" into "one window at a time, second caller is told no." Which one is right depends on what habluetooth's scheduler expects — see the inline suggestion about serialization-vs-coalescing on the same block. If habluetooth's contract is "if you return True I trust the device is active for `duration` seconds starting now," then returning True from a queued caller is actually the buggier behavior, and fail-fast is the safer default.

aioshelly/ble/backend/scanner.py
bluetoothbot: Agreed this is followup-sized. The circular import is real: `aioshelly.ble.__init__` imports from `aioshelly.ble.backend.scanner` (for `create_scanner`), and pulling `async_start_scanner` up to module level here would close the loop. The cleanest unwind is probably to move `async_start_scanner` (and the scripts it provisions) into a leaf module like `aioshelly.ble.script` that neither `__init__` nor `scanner` need to import eagerly, then re-export it from `aioshelly.ble` for back-compat. That's a non-trivial refactor — happy to do it in a separate PR once this one lands and we know the API shape is stable.

As a smaller intermediate step: the late import could be hoisted to module scope as `from aioshelly.ble import async_start_scanner as _ble_async_start_scanner` inside a `TYPE_CHECKING`-style lazy guard, but honestly the current `noqa: PLC0415` inside the method is the lowest-risk option until the refactor happens.

aioshelly/ble/backend/scanner.py
bluetoothbot: The `if self._active_window_lock.locked(): return False` check at the top of the method (lines 61-62 in the diff) is doing double duty as your reentrancy guard. `asyncio.Lock` is not reentrant — a same-task re-acquire would silently deadlock on `acquire()` — so the `locked()` fail-fast is what saves you from the same-task reentry case as well as the concurrent-task overlap case. The only path that could still surprise you is a same-task call from *inside* the `async with` block (e.g., a callback wired into `_ble.async_start_scanner` that loops back through the scanner), which would also return False here — arguably the correct outcome since the device is already in active mode for the in-flight window. So I'd argue the guard you want is already in place; the only remaining question is whether habluetooth's scheduler ever calls back into `async_request_active_window` from its own callback chain, and if so whether you'd prefer False or coalescing — but the bot's earlier reply already covered that design choice.

aioshelly/ble/backend/scanner.py
bluetoothbot: Worth noting that the current diff has `from aioshelly import ble as _ble` at module top (line 12), not as a late import — which is what closes the circular loop with `aioshelly/ble/__init__.py:13`'s `from .backend.scanner import ShellyBLEScanner`. It happens to work because `_ble.async_start_scanner` is only dereferenced inside the method (after both modules finish loading), but it's brittle: any future change that touches `_ble.<attr>` at module or class scope (default arg, decorator, class attribute) will raise `AttributeError` at import time, since at the point `scanner.py` first sees `_ble` the `aioshelly.ble` module is still partially initialized and `async_start_scanner` isn't bound yet.

The followup the bot sketched earlier — extracting `async_start_scanner` and the script constants into a leaf module like `aioshelly.ble.script` that neither `__init__` nor `scanner` imports eagerly, then re-exporting from `aioshelly.ble` for back-compat — is the right shape and keeps the public API untouched. Agree it's not blocking for this PR; flagging in case the import order changes before that refactor lands.

aioshelly/ble/backend/scanner.py
```diff
@@ -2,21 +2,106 @@
...
+        device = self._active_window_device
+        event_type = self._active_window_event_type
+        data_version = self._active_window_data_version
+        if device is None or event_type is None or data_version is None:
+            return False
+        if self._active_window_lock.locked():
+            return False
+
+        async with self._active_window_lock:
+            try:
```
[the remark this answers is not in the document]
bdraco: seems overkill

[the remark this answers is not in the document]
bdraco: that can't actually happen with asyncio

aioshelly/ble/backend/scanner.py
bluetoothbot: You already have the reentrancy guard you're describing: the `if self._active_window_lock.locked(): return False` at the top of the method does it. Two reasons it's sufficient even for a public API:

1. asyncio.Lock is non-reentrant, so a same-task re-entry would deadlock on `acquire()` rather than misbehave. The `locked()` fast-fail converts that into a clean `False` before you ever reach `async with`.
2. It's race-free against concurrent tasks too — there is no `await` between `locked()` returning False and the `async with` acquiring the lock, so in single-threaded asyncio no other coroutine can interleave and slip in. The check-then-acquire is effectively atomic.

So both the concurrent-caller case and the same-task case are covered. The only thing left to confirm with habluetooth is the *semantics* of returning `False` on overlap: the current code is fail-fast ("one window at a time, second caller told no"), which is the right default if habluetooth treats a `True` return as "the device is active for `duration` starting now." If habluetooth instead expects coalescing/extension, you'd want to extend the in-flight window rather than reject — but fail-fast is the safer of the two if the contract is ambiguous.

aioshelly/ble/backend/scanner.py
bluetoothbot: Agreed this is followup-sized. One clarification on the current state: the diff already hoisted this to a module-top import (`from aioshelly import ble as _ble`, line 12), not a late import inside the method. It resolves only because `aioshelly.ble` is already in `sys.modules` (partially initialized) when `scanner.py` is imported via `aioshelly/ble/__init__.py:13`, and `_ble.async_start_scanner` is dereferenced lazily at call time after both modules finish loading. It works, but it's brittle — any future top-level use of `_ble.<symbol>` during import would break the cycle.

The clean unwind is to extract `async_start_scanner` (and the script-provisioning helpers it uses) into a leaf module like `aioshelly.ble.script` that neither `ble/__init__` nor `scanner` need to import eagerly, then re-export from `aioshelly.ble` for back-compat. That breaks the cycle without the lazy-binding dance and lets you import the symbol directly. Reasonable to defer to a separate PR once this API shape stabilizes.

aioshelly/ble/backend/scanner.py
```diff
@@ -2,21 +2,89 @@
...
 import logging
-from typing import Any
+from typing import TYPE_CHECKING, Any
 
 from bluetooth_data_tools import monotonic_time_coarse
 from habluetooth import BaseHaRemoteScanner
 
+from aioshelly import ble as _ble
+
+from ...exceptions import ShellyError
```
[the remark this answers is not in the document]
bdraco: just tested, not true

aioshelly/ble/backend/scanner.py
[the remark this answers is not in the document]
bluetoothbot: On the single-cancellation path your `test_..._restore_runs_under_cancellation` exercises, the `asyncio.shield` is in fact redundant: when the task is cancelled during `await asyncio.sleep(duration)`, the `CancelledError` is delivered once at the sleep, and the `finally` then runs with no cancellation pending — so a bare `await _ble.async_set_active_mode(device, active=False)` would already run to completion before the `CancelledError` resumes propagating. The test passes with or without the shield, so it doesn't actually prove the shield is load-bearing.

Where it's *not* overkill is double-cancellation: if a second `cancel()` lands while you're awaiting the restore (e.g. an `asyncio.wait_for`/`timeout` firing on top of an explicit cancel, or supervisory teardown), a bare `await` would have its restore interrupted and the device could stay active. The shield keeps the restore coroutine running in that case. So: keep the `try/finally` regardless — that's the part that guarantees the restore on the normal cancel path — and treat the `shield` as cheap insurance against the double-cancel corner. Drop it only if you're confident habluetooth never double-cancels the window task. (See my reply on the sibling comment for one caveat about what shield does *not* guarantee.)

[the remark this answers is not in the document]
bluetoothbot: Worth pinning down exactly what can and can't happen. Cancellation mid-window absolutely *can* happen — anything holding the task can cancel it (config-entry unload, a wrapping `wait_for` timeout, scheduler teardown), so the `finally` is doing real work, not guarding a phantom. What asyncio guarantees: the `finally` always runs, and for a *single* `cancel()` the restore inside it completes, because the `CancelledError` was already delivered at the `sleep` and the next `await` is therefore not itself cancelled.

The subtlety is what `asyncio.shield` does *not* give you: if the outer task is cancelled *again* while awaiting the shielded restore, the shielded coroutine keeps running but the `await` returns immediately via `CancelledError` — the function does not block until the restore finishes. The device still reverts, but as a detached task after this coroutine has unwound, so 'shield guarantees passive before we return' isn't strictly true on a double-cancel. If you want a hard before-return guarantee you'd catch `CancelledError` around the sleep and `await` the restore unshielded; if best-effort revert is acceptable (it usually is for a battery safeguard), the current shape is fine. Net: the scenario isn't impossible, but the bare `try/finally` already covers the realistic single-cancel case, so simplifying away the shield is a reasonable call if you'd rather not reason about double-cancel.

aioshelly/ble/backend/scanner.py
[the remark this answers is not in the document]
bluetoothbot: Agreed, it works — and it's worth capturing *why*, because the pattern is load-bearing and easy to break by accident later.

`from aioshelly import ble as _ble` binds the **module object**, not a name pulled out of it. The attributes (`_ble.async_get_ble_script_id`, `_ble.async_set_active_mode`) are only dereferenced at call time, long after both modules have finished importing. During `aioshelly/ble/__init__.py`'s line-13 `from .backend.scanner import ShellyBLEScanner`, this file executes and re-enters the partially-initialized `aioshelly.ble` via the `sys.modules` fallback — which succeeds *precisely because* you're importing the module, not `async_set_active_mode` (which isn't bound yet at that point in `__init__`). Importing `aioshelly.ble.backend.scanner` directly is also safe: Python fully executes the parent `aioshelly.ble` package before the submodule reference resolves.

The one edit that *would* break it on the circular path: switching this to `from aioshelly.ble import async_set_active_mode` at module top — that's a name import against a not-yet-defined attribute and raises `ImportError`. So the late-import followup is genuinely optional; the current form is correct as-is. Suggest just adding a one-line comment ('module-reference import, not a name import — keep it that way to avoid the circular ImportError') so a future cleanup pass doesn't 'tidy' it into the broken form.

bdraco:
Remaining bot comments don't make sense to change on based on habluetooth design

aioshelly/ble/__init__.py
```diff
@@ -75,6 +75,36 @@ async def async_start_scanner(
...
+async def async_get_ble_script_id(device: RpcDevice) -> int | None:
+    """Return the installed BLE integration script id, or None if absent.
+
+    Callers that toggle active mode repeatedly (e.g. the active-window
+    scheduler) should resolve the id once and cache it for the
+    lifetime of the script, since the id is stable until the script
+    is recreated.
+    """
+    script_name_to_id = await _async_get_scripts_by_name(device)
+    return script_name_to_id.get(BLE_SCRIPT_NAME)
```
chemelli74, line 87: Do we want to check also script version to be sure the new one is installed ?

aioshelly/ble/__init__.py
```diff
@@ -75,6 +75,36 @@ async def async_start_scanner(
...
+    in const.py) that stops and restarts BLE.Scanner with the new
+    active flag. Driving the toggle via Script.Eval keeps the script
+    body byte-identical across windows, so the device's flash is not
+    rewritten on every transition.
+
+    ``script_id`` must be the id returned by ``async_get_ble_script_id``
+    or otherwise known to the caller. This keeps each flip a single
+    Script.Eval round-trip with no Script.List overhead.
+    """
+    await device.script_eval(script_id, f"setActive({'true' if active else 'false'})")
```
chemelli74, line 105: Should we guard for old version script ?

chemelli74 (approved):

PR #1154, opened 2026-05-25
Reflect active scan windows in ShellyBLEScanner.current_mode

bdraco:
When the auto scheduler flips a Shelly proxy into an active scan window the device gets the right `Script.Eval`, but `self.current_mode` is never updated, so the manager is never notified and the UI keeps showing the proxy as passive even while it is sweeping actively.

The override now calls `set_current_mode(ACTIVE)` after the entry flip succeeds and restores the prior value in the finally block, matching how the local `HaScanner` tracks its own window; `requested_mode` is untouched, only `current_mode` reflects the live state.

Bumps habluetooth to >=6.3.0

### DIFF
--- a/aioshelly/ble/backend/scanner.py
+++ b/aioshelly/ble/backend/scanner.py
@@ -85,21 +85,31 @@ class ShellyBLEScanner(BaseHaRemoteScanner):
                 await asyncio.sleep(duration)
             finally:
                 try:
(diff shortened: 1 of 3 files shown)


thecode (approved):

chemelli74 (approved):

bdraco:
thanks

PR #1163, opened 2026-06-02
Avoid double-decoding RPC WebSocket frames with aiohttp decode_text

bdraco:
aiohttp 3.14 added `decode_text` to `ws_connect`. Passing `decode_text=False` delivers TEXT frames as raw bytes, so orjson decodes them directly instead of decoding to str first and re-scanning. The send path already used `send_frame(json_bytes(...))`, so this finishes the round trip.

Also dropped `aioresponses` for a small local mock in the common tests. No aioresponses release works with aiohttp 3.14 yet, and it was overkill for what these tests need.


bdraco:
@bluetoothbot review

thecode (approved):

PR #1213, opened 2026-07-20
Use uv as package manager

bieniu:


.devcontainer/Dockerfile
```diff
@@ -16,5 +16,7 @@ RUN if [ "${INSTALL_NODE}" = "true" ]; then su vscode -c "umask 0002 && . /usr/l
 # [Optional] Uncomment this line to install global node packages.
 # RUN su vscode -c "source /usr/local/share/nvm/nvm.sh && npm install -g <your-package-here>" 2>&1
 
+RUN pip3 install uv
```
bieniu, line 19: Let `uv` be available from devcontainer

.pre-commit-config.yaml
```diff
@@ -3,7 +3,7 @@ repos:
     rev: v3.21.2
     hooks:
       - id: pyupgrade
-        args: ["--py310-plus"]
+        args: ["--py311-plus"]
```
bieniu, line 6: We support 3.11+

pyproject.toml
```diff
@@ -31,7 +31,7 @@ dependencies = [
   "zeroconf>=0.148.0"
 ]
 
-[project.optional-dependencies]
+[dependency-groups]
```
bieniu, line 34: Migrating from `project.optional-dependencies` to `dependency-groups` because the former publishes development tools as installable PyPI extras (unnecessary metadata for end users, currently `pip install aioshelly[dev]` works) while the latter keeps dev/lint dependencies local to the project and enables `uv sync --all-groups` without polluting the published package interface.

pyproject.toml
```diff
@@ -113,26 +114,28 @@ skip_missing_interpreters = True
 
 [gh-actions]
 python =
-  3.11: py311, lint, prek, ty
+  3.11: py311, tests, lint, prek, ty
```
bieniu, line 121: Previously we didn't run tests for Py3.11. I'm not sure if that was intentional.

thecode (approved):

PR #1218, opened 2026-07-22
Add RPC cb_set method

thecode:
Add support for controlling the circuit breaker (CB) component.

API: https://shelly-api-docs.shelly.cloud/gen2/ComponentsAndServices/CB#cbset

Note: API has incorrect documentation - output accept true if the breaker is not in a safety limit, this has reported and discussed with Shelly and they will update the docs.

**Turn on:**
`http://192.168.1.47/rpc/CB.Set?id=0&output=true`

**Turn on response:**
`{"was_on":false}`

**Turn off:**
`http://192.168.1.47/rpc/CB.Set?id=0&output=false`

**Turn off response:**
`{"was_on":true}`

### DIFF
--- a/aioshelly/rpc_device/device.py
+++ b/aioshelly/rpc_device/device.py
@@ -563,6 +563,14 @@ class RpcDevice:
         }
         await self.call_rpc("Switch.Set", params=params)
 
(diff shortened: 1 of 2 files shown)


bieniu (approved):
Looks good, thanks

PR #1225, opened 2026-07-24
Use hashes instead of version numbers for GH Actions

bieniu:


thecode (approved):

chemelli74:
I think we can switch to renovate, that will take care of those kind of improvements over time.

chemelli74 (approved):

PR #1241, opened 2026-08-13
Fix Wall Display X2i model identifier

sasparillax:
Remove an unintended leading whitespace from the Shelly Wall Display X2i model identifier.

The device reports SAWD-5A1XX10EU0, while the current constant contains a leading space:

" SAWD-5A1XX10EU0"

This prevents Home Assistant from recognizing the X2i as a Wall Display, so the screen on/off button entities are not created.

Tested with Home Assistant Core 2026.8.1 and a Shelly Wall Display X2i. After removing the leading whitespace and restarting Home Assistant Core, both screen on/off buttons appeared correctly.


thecode (approved):
Thanks!

PR #1249, opened 2026-08-18
Add methods for camera switches

bieniu:
https://preview-api-docs.shelly.cloud/502-camaera-add-device-api-docs/gen2/ComponentsAndServices/Camera/#cameraset


aioshelly/const.py
```diff
@@ -203,7 +203,7 @@
 GEN4_MIN_FIRMWARE_DATE = 20240902
 
 # Firmware with RTSP support
-GEN3_CAMERA_MIN_FIRMWARE_DATE = 20260724
+GEN3_CAMERA_MIN_FIRMWARE_DATE = 20260817
```
bieniu, line 206: This firmware moves `arm` and `privacy` from config to status.

thecode (approved):

PR #1252, opened 2026-08-22
Use aiohttp.encode_basic_auth instead of deprecated BasicAuth

balloobbot:
`aiohttp.BasicAuth` and the request `auth` parameter are deprecated and will be removed in aiohttp 4.0. Using `aiohttp.encode_basic_auth()` with an `Authorization` header now makes that migration a non-event.

aiohttp is already pinned to >=3.14.0, which is the version that added `encode_basic_auth()`.

`ConnectionOptions.auth` becomes `ConnectionOptions.auth_header` and now holds the header value. It is still set from `username`/`password`, so callers that pass credentials are unaffected.

I'm a bot, asked by @balloob to fix this because the deprecation warning shows up in Home Assistant CI.

### DIFF
--- a/aioshelly/block_device/device.py
+++ b/aioshelly/block_device/device.py
@@ -162,7 +162,7 @@ class BlockDevice:
                 self.aiohttp_session, self.ip_address, self.options.device_mac
             )
 
-            if self.requires_auth and not self.options.auth:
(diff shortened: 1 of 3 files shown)


thecode (approved):
Tested with Shelly 2.5

PR #1261, opened 2026-08-25
Add `ir_emit_raw` method

bieniu:
## `IR.EmitRaw` — Detailed API
 
On-demand raw IR transmit. Sends a pulse list passed **at call time** — no
learning, no stored component. This is the method HA's IR emitter integration
targets.
 
### Request
 
`POST http://<device>/rpc/IR.EmitRaw`
 
```json
{
  "timings": [4500, -4500, 560, -1690, 560, -560, 560, -40000],
  "freq": 38000,
  "repeats": 0
}
```
 
### Parameters
 
| Field | Type | Required | Default | Range | Description |
|-------|------|----------|---------|-------|-------------|
| `timings` | array of signed int | yes | — | 1..`kMaxRawTimings` (1024) | Pulse durations in **microseconds**. **`+` = mark** (carrier/LED on), **`−` = space** (off). Values alternate mark/space and are packed two-per `rmt_symbol_word_t`. Odd length → the last symbol's `duration1` is 0 (end marker). |
| `freq` | int | no | `38000` | `kMinCarrierHz`..`kMaxCarrierHz` (20000..60000) | Carrier frequency in Hz. Applied via `IrDriver::SetCarrierFreq`, which re-applies the carrier only when it changes (e.g. 40000 for Sony). |
| `repeats` | int | no | `0` | 0..`kMaxRepeats-1` | **Extra** repeats. `0` = send once. Total transmissions = `repeats + 1`. |

### DIFF
--- a/aioshelly/rpc_device/device.py
+++ b/aioshelly/rpc_device/device.py
@@ -683,6 +683,11 @@ class RpcDevice:
         params = {"id": id_, "config": {"rtsp": {"enable": value}}}
         await self.call_rpc("Camera.SetConfig", params)
 
+    async def ir_emit_raw(self, timings: list[int], freq: int, repeats: int) -> None:
+        """Emit raw IR data."""
+        params = {"timings": timings, "freq": freq, "repeats": repeats}
+        await self.call_rpc("IR.EmitRaw", params)
+
     async def poll(self) -> None:
         """Poll device for calls that do not receive push updates."""
         calls: list[tuple[str, dict[str, Any] | None]] = [("Shelly.GetStatus", None)]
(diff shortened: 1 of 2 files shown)


aioshelly/rpc_device/device.py
```diff
@@ -683,6 +683,11 @@ async def set_camera_rtsp(self, id_: int, value: bool) -> None:
         params = {"id": id_, "config": {"rtsp": {"enable": value}}}
         await self.call_rpc("Camera.SetConfig", params)
 
+    async def ir_emit_raw(self, timings: list[int], freq: int, repeats: int) -> None:
```
chemelli74: Would it make sense to have a `repeats` default to 1 ?

bieniu: The default value for `repeats` is `0` and tbh I was thinking of adding it, just in case. Same, for `freq`- `38000`.

WDYT?

chemelli74: Go ahead

thecode (approved):

thecode (approved):

Issue #1230, opened 2026-07-27
Shelly Plug S Gen3 firmware 2.0.0: opaque digest-auth nonce appears to defeat proactive re-auth — exactly one failed poll per nonce lifetime, per device


rs-com:
### Problem

On Shelly Plug S Gen3 devices running firmware `2.0.0`, every authenticated RPC polling cycle fails **exactly once per ~62 minutes**, per device, and recovers on the very next poll 60 s later:

```
ERROR [homeassistant.components.shelly] Error fetching CPO2 data: An error occurred while retrieving data from CPO2
DEBUG [homeassistant.components.shelly] Finished fetching CPO2 data in 0.029 seconds (success: False)
...
INFO  [homeassistant.components.shelly] Fetching CPO2 data recovered
DEBUG [homeassistant.components.shelly] Finished fetching CPO2 data in 0.082 seconds (success: True)
```

Identical devices on firmware `1.7.5` in the same installation, on the same Wi-Fi AP, **never** fail. Over ~25 days this produced **1030 `Error fetching …` log lines** from 5 devices.

Note the failure takes **0.029 s** — it is an immediate rejection, not a timeout.

### Root cause evidence: the nonce format changed

Comparing the `send(...)` debug lines of a *successful* request on each firmware (this was the decisive clue — the error message itself carries no information):

**Firmware 1.7.5 — never fails**
```
send(<ip>:80): {'id': 1953, 'method': 'Shelly.GetStatus',
  'auth': {'realm': 'shellyplugsg3-<redacted>', 'username': 'admin',
           'nonce': 1785145983,          <-- integer == Unix timestamp of issuance
           'nc': 62,                     <-- counter climbs freely, 1/min
           'cnonce': '...', 'response': '...', 'algorithm': 'SHA-256'}}
```

**Firmware 2.0.0 — fails once per ~62 min**
```
send(<ip>:80): {'id': 1772, 'method': 'Shelly.GetStatus',
  'auth': {'realm': 'shellyplugsg3-<redacted>', 'username': 'admin',
           'nonce': 'd0Vnag...<opaque base64, 64 chars>...',   <-- opaque string
           'nc': 3,                      <-- resets to a low value after each failure
           'cnonce': '...', 'response': '...', 'algorithm': 'SHA-256'}}
```

On 1.7.5 the nonce **is** its own issuance timestamp. On 2.0.0 it is an opaque token from which no age can be derived.

[...]

PR #1263, opened 2026-08-28
Drop the cached auth nonce when the device answers 429

loeffelsteve:
Problem
_raise_for_unrecoverable_errors() refreshes the auth challenge only for 401. Every other code, including 429, raises immediately and leaves AuthData untouched.

Per the [Gen2 authentication docs](https://shelly-api-docs.shelly.cloud/gen2/General/Authentication/), a device counts failed authentications in a 10-minute sliding window and, once throttled, answers 429 instead of 401 — including for requests that would otherwise hand out a fresh nonce. So exactly when a new challenge is needed, the code path that would obtain one is unreachable.

That turns a transient condition into a persistent one. The cached nonce is dead, every following call resends it with only nc incremented, each one counts as another failed authentication, and the window is continuously re-armed. A device whose coordinator keeps polling never falls silent long enough for the throttle to drain.

Observed on a device in this state: the same nonce went out with nc climbing 2 → 53 without a single refresh attempt, while a separate WebSocket client authenticated and executed Switch.Set against that same device successfully at that moment. Reloading the config entry did not help either, since AuthData survives it — from the user's side the device is simply gone.

Change
On 429, drop the cached nonce. The next request then goes out unauthenticated and picks up a fresh challenge through the existing retry path, instead of replaying one the device has already rejected. The call itself still raises RpcCallError, so nothing changes for callers.

This is deliberately minimal: no retry loop and no timer. It only removes the state that prevents recovery. Adding a real backoff — not sending anything for a while after a 429, since requests inside the window extend it, worst case 5 minutes — would be a sensible follow-up, but it needs new state and is a larger discussion.

Context
The firmware side that made this visible is fixed in Shelly 2.0.1-beta1 (see [home-assistant/core#177257](https://github.com/home-assistant/core/issues/177257), and #1230 for the matching once-per-nonce-lifetime poll failure). This change is about the library's own behaviour, which stays a robustness gap regardless of firmware: any future throttle wedges the connection the same way.

Also worth considering separately: RpcCallError's code and message are dropped before the error reaches the user. In Home Assistant, components/shelly/entity.py maps every RpcCallError to rpc_call_action_error with only the entity and device name, so the log line gives no hint that an authentication throttle is involved. Surfacing the device's code and message would make this class of problem far easier to recognise.

### DIFF
--- a/aioshelly/rpc_device/wsrpc.py
+++ b/aioshelly/rpc_device/wsrpc.py
@@ -461,7 +461,7 @@ class WsRPC(WsBase):
         except KeyError as err:
             raise RpcCallError(0, f"bad response: {resp}") from err
 
-                if code != HTTPStatus.UNAUTHORIZED.value:
+        if code != HTTPStatus.UNAUTHORIZED.value:
             if (
                 code == HTTPStatus.TOO_MANY_REQUESTS.value
                 and (auth_data := self._session.auth_data) is not None
--- a/aioshelly/rpc_device/wsrpc.py
+++ b/aioshelly/rpc_device/wsrpc.py
@@ -461,7 +461,17 @@ class WsRPC(WsBase):
         except KeyError as err:
             raise RpcCallError(0, f"bad response: {resp}") from err
 
-        if code != HTTPStatus.UNAUTHORIZED.value:
+                if code != HTTPStatus.UNAUTHORIZED.value:
+            if (
+                code == HTTPStatus.TOO_MANY_REQUESTS.value
+                and (auth_data := self._session.auth_data) is not None
+            ):
+                # A device that throttles repeated failed authentications answers
+                # 429 instead of issuing a new 401 challenge, so the cached nonce
+                # can no longer be refreshed through the usual path. Replaying it
+                # makes every following call fail as well, so drop it and let the
+                # next call solicit a fresh challenge.
+                auth_data.nonce = ""
             raise RpcCallError(code, msg)
 
         if allow_auth_retry and self._session.auth_data is not None:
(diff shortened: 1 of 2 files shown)


aioshelly/rpc_device/wsrpc.py
```diff
@@ -462,6 +462,16 @@ def _raise_for_unrecoverable_errors(
...
+            if (
+                code == HTTPStatus.TOO_MANY_REQUESTS.value
+                and (auth_data := self._session.auth_data) is not None
+            ):
+                # A device that throttles repeated failed authentications answers
+                # 429 instead of issuing a new 401 challenge, so the cached nonce
+                # can no longer be refreshed through the usual path. Replaying it
+                # makes every following call fail as well, so drop it and let the
+                # next call solicit a fresh challenge.
+                auth_data.nonce = ""
```
bieniu, line 474: I would add some debug logging here.

thecode: As I explained on https://github.com/home-assistant/core/issues/177257 firmware 2.0.0 has a bug, setting `nonce` to an empty string will not fix or change anything as implemented in this PR because we won't update again.

On a normal device we should never reach "Too many requests" and if we do there is a serious error and we should not try to recover.

I would prefer that if we catch "Too many requests" we raise a specific error, but I don't know what to do with this error.

Please first explain what you are trying to fix and show a log what it fixes.

thecode (changes requested):

loeffelsteve:
Closing this — I can't produce the log you asked for, and the field data now argues against the change.

Ten days of measurements from a 56-device installation, after moving the two worst-affected devices to firmware 2.0.1-beta1:

firmware	switch operations	RPC errors
2.0.1-beta1	229	0
2.0.0	586	9
1.7.5	~450	0
The two devices now at zero were previously the worst in the installation (123 and 68 errors). The remaining errors occur exclusively on devices still running 2.0.0. I have since rolled 2.0.1-beta1 out to all 23 reachable Gen2/Gen3 devices here.

On the substance of your objection: my premise that a polling client keeps the throttle armed indefinitely was asserted, not measured — and my own capture contradicts it, since Home Assistant recovered on its own once the window elapsed. Without that premise the change has no demonstrable benefit, so closing is the right call. Thanks both for the review.

PR #1284, opened 2026-09-13
Add `camera_get_image` method

bieniu:
https://shelly-api-docs.shelly.cloud/gen2/ComponentsAndServices/Camera#get-cameraidsnapshot


aioshelly/rpc_device/device.py
```diff
@@ -665,6 +670,39 @@ async def media_list_radio_stations(self) -> list[dict[str, Any]]:
         result = await self.call_rpc("Media.Radio.ListFavourites")
         return result["list"]
 
+    async def camera_get_image(self, camera_id: int) -> bytes:
+        """Return a still image from the camera's HTTP snapshot endpoint."""
+        if TYPE_CHECKING:
+            assert self.aiohttp_session
+
+        middlewares = None
```
chemelli74:
```suggestion
        middlewares DigestAuthMiddleware | None = None
```

aioshelly/rpc_device/device.py
```diff
@@ -665,6 +670,39 @@ async def media_list_radio_stations(self) -> list[dict[str, Any]]:
...
+        if self.options.username and self.options.password:
+            middlewares = (
+                DigestAuthMiddleware(self.options.username, self.options.password),
+            )
+
+        try:
+            async with self.aiohttp_session.get(
+                URL.build(
+                    scheme="https" if use_ssl(self.port) else "http",
```
chemelli74, line 687: why a different check here and a few lines below ?

`use_ssl(self.port)` vs `self.options.verify_ssl`

bieniu: The `ssl` argument in the aiohttp actually means "verify ssl" so we can't use `"https" if use_ssl(self.port) else "http"` for it.

obraz

aioshelly/rpc_device/device.py
```diff
@@ -665,6 +670,39 @@ async def media_list_radio_stations(self) -> list[dict[str, Any]]:
...
+            ) as resp:
+                if resp.status == HTTPStatus.UNAUTHORIZED:
+                    raise InvalidAuthError(resp.status)
+                if resp.status != HTTPStatus.OK:
+                    raise HttpCallError(
+                        resp.status, f"Snapshot endpoint returned HTTP {resp.status}"
+                    )
+                return await resp.read()
+        except CONNECT_ERRORS as err:
+            raise DeviceConnectionError(err) from err
```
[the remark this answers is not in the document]
thecode: Copilot note make sense, we modified in the past all places so that for Timeout we raise a different error so we can show it differently than other errors.
https://github.com/home-assistant-libs/aioshelly/blob/ca3be12f8ab9cc2fd2d6ae1b4f0c33a80b782502/aioshelly/common.py#L121-L124

thecode (approved):

